PR #39, opened 2023-04-19
Use middleware instead Rails controller for model query

st0012:
- By using middleware instead of a Rails controller, it skips many middlewares that aren't needed for our model query request. In big Rails apps, this could be a big performance improvement.
- Although Rails engine also supports being used as a middleware, it still requires registering routes to the application. That doesn't always work depends on the app's own rules on how to register routes.
- Compare to Rails engine, middleware is a more flexible way to integrate with Rails apps and also a lot lighter.

Closes https://github.com/Shopify/ruby-dev-exp-issues/issues/898


test/lib/middleware_test.rb
```diff
@@ -6,10 +6,8 @@
 module RubyLsp
   module Rails
     class ModelsControllerTest < ActionDispatch::IntegrationTest
-      T.unsafe(self).include(Engine.routes.url_helpers)
-
       test "GET show returns column information for existing models" do
-        get model_url(model: "User")
```
andyw8, line 12: You should be able to remove the RBI shim for this.

lib/ruby_lsp_rails/middleware.rb
```diff
@@ -0,0 +1,63 @@
...
+      end
+
+      sig { params(env: T::Hash[T.untyped, T.untyped]).returns(T::Array[T.untyped]) }
+      def call(env)
+        request = ActionDispatch::Request.new(env)
+        # TODO: improve the model name regex
+        match = request.path.match(PATH_REGEXP)
+
+        if match
+          resolve_database_info_from_model(match[:model_name])
```
andyw8, line 23: nit: We could return early here, then change `rescue` to `ensure`, that way we only need `@app.call(env)` once.

lib/ruby_lsp_rails/middleware.rb
```diff
@@ -0,0 +1,63 @@
+# typed: strict
+# frozen_string_literal: true
+
+module RubyLsp
+  module Rails
+    class Middleware
```
andyw8, line 6: We might later want to consider extracting some of this into `ruby-lsp`, so there's less 'boilerplate' necessary for extension authors.

st0012:
> I'll try it out shortly, but I assume that since this is done at the rack level, we'll no longer see any /ruby_rails_lsp/... requests in the Rails server log?

The middleware is not injected that early in the stack, so you should still see it in logs, like:

`Started GET "/ruby_lsp_rails/models/User" for 127.0.0.1 at 2023-04-20 11:56:51 +0100`

lib/ruby_lsp_rails/middleware.rb
st0012: But if we won't want them to do this, what's the point of extracting them to `ruby-lsp`?

This is Rack/Rails specific pattern and not what `ruby-lsp` should have provided?

vinistock: I agree with Stan here, `ruby-lsp-rails` will be quite a unique extension since it will be communicating with a live server. I suspect most other extensions wouldn't need it - and we shouldn't create incentive for people to do so as getting this wrong can be detrimental to the overall experience.

andyw8: I think there there could be other neat extensions powered by a similar mechanism, but I agree with not trying to optimize for that until we actually have more real-world examples.

PR #59, opened 2023-04-28
Add code lens extension (for Rails tests)

adisonlampert:
⚠️ Must be merged after https://github.com/Shopify/ruby-lsp/pull/666

### Description 

Add code lens extension to recognize Rails tests:

```ruby
test "should test something" do
    # assert something
end
```

A Rails test is a `SyntaxTree::Command` with the test name as the first argument so the extension implements `on_command` to create code lenses for Rails tests.

https://github.com/Shopify/ruby-lsp/pull/666 adds parameters to `initialize` so I also updated the hover extension + tests.

### Testing

Created `code_lens_test.rb` that checks that Rails-styled tests are recognized and a `CodeLens` response is created.

You can also manually test by checking out this branch and running the tests in `code_lens_test.rb` from the file.
- Note: Until https://github.com/Shopify/ruby-lsp/pull/666 is merged, you must make sure the `ruby-lsp` gem is installed from [this branch](https://github.com/Shopify/ruby-lsp/tree/al/refactor-code-lens)

### DIFF
--- a/lib/ruby_lsp/ruby_lsp_rails/code_lens.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/code_lens.rb
@@ -1,0 +1,74 @@
+# typed: strict
+# frozen_string_literal: true
+
+module RubyLsp
+  module Rails
+    class CodeLens < ::RubyLsp::Listener
+      extend T::Sig
+      extend T::Generic
+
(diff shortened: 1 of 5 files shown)


andyw8:
I have continued this in https://github.com/Shopify/ruby-lsp-rails/pull/67 - I'm leaving this as-is to allow for a comparison.

Issue #47, opened 2023-04-21
Check if we can depend on `railties` instead of `rails`


st0012:
Now this gem is not a Rails engine anymore, we can probably just depend on `railties` instead of `rails`.
The reason is that:
1. In general, the smaller the dependency the better.
2. In the past I encountered some apps that don't have `rails` in Gemfile but only individual component gems (because for example, they don't use `active_storage`). And they should still be use this gem without bundling the gems they don't rely on.

PR #60, opened 2023-04-30
Only require parts of Rails actually needed

bdewater:
In our app we don't need Action Cable/Text/Mailbox etc and we don't want to pull it in for this gem.

It would be nice to even make Active Record optional and have a pattern to only offer functionality for Rails components used in the consuming application.

Relates to #47 but is more of a quick fix - I think it's possible to depends on Railties alone but this solves my immediate problem.


bdewater:
@vinistock sorry for the late response, I was on pat leave :). I'll rebase this later this week.

Having to re-add parts of Rails explicitly to me felt like a good thing, so devs do not unintentionally add functionality relying on parts of the framework not declared as  dependencies.

rafaelfranca:
I agree wit Bart, we should not add Rails in the Gemfile.

bdewater:
Rebased! Seems the CLA check is broken.

st0012:
@bdewater some dependencies seem downgraded in this PR and CI is failing due to dependency issues. Can you resolve it and push again?

st0012 (approved):
Thanks for the PR 👍

PR #64, opened 2023-05-08
Update RBIs

andyw8:
Needed for https://github.com/Shopify/ruby-lsp-rails/pull/59

Ran `bin/tapioca gem`


andyw8:
(I had initially includes the RBI for Psych, which wasn't present previously but is added by `bin/tapioca gem`. It introduces a lot of type errors, I didn't yet look into why).

PR #87, opened 2023-06-22
Ensure `app_uri.txt` is cleaned up when Rails server shuts down

andyw8:
The `app_uri.txt` file should only exist when the server is running. The extension uses its presence to report if the server is running or not. If the server is not running, some of the extension features will not be available.

### DIFF
--- a/lib/ruby_lsp_rails/railtie.rb
+++ b/lib/ruby_lsp_rails/railtie.rb
@@ -14,7 +14,15 @@ module RubyLsp
           if defined?(::Rails::Server)
             ssl_enable, host, port = ::Rails::Server::Options.new.parse!(ARGV).values_at(:SSLEnable, :Host, :Port)
             app_uri = "#{ssl_enable ? "https" : "http"}://#{host}:#{port}"
-            File.write("#{::Rails.root}/tmp/app_uri.txt", app_uri)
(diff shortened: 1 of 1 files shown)


st0012 (approved):
👍

vinistock:
I'm not sure I understand. We check if the server is running by [making a request to it](https://github.com/Shopify/ruby-lsp-rails/blob/62cbae4e7e65703a52986066ca1fe953bd583aff/lib/ruby_lsp/ruby_lsp_rails/rails_client.rb#L62).

But we need to ensure that `app_uri.txt` exists because otherwise we don't know the URL we need to make requests against.

The extension currently [complains if app_uri.txt doesn't exist](https://github.com/Shopify/ruby-lsp-rails/blob/62cbae4e7e65703a52986066ca1fe953bd583aff/lib/ruby_lsp/ruby_lsp_rails/rails_client.rb#L35) because we wouldn't know which URL to make requests against.

The idea of that error was for it to show up once and then after restarting the server the file would exist and you'd never need to do it again.

st0012:
As long as our implementation still relies on the file's presence as a state, either for error like now or just a warning like in #88, we should make sure the state doesn't leak to another run.

If we don't do things based on `File.exist?(app_uri_path)` anymore, I'm fine with leaving it uncleaned.

PR #67, opened 2023-05-08
Add code lens extension (for Rails)

andyw8:
Based on https://github.com/Shopify/ruby-lsp-rails/pull/59

Reminder: Ensure @adisonlampert is credited in the release notes.

This depends on https://github.com/Shopify/vscode-ruby-lsp/pull/639

TODO:
- [x] Get working with Test Explorer
- [x] Verify Debug CodeLens
- [x] Update docs
- [x] Add screenshot gif
- [x] Specify minimum Ruby LSP version (needs https://github.com/Shopify/ruby-lsp/pull/734 to be released). Also remove the Gemfile entry.
- [x] Update types (avoid `T.unsafe`)

### Testing

```
  gem "ruby-lsp", github: "Shopify/ruby-lsp"
  gem "ruby-lsp-rails", github: "Shopify/ruby-lsp-rails", branch: "andyw8/add-code-lens-extension"
```

PR #88, opened 2023-06-22
Change how we warn about server not running

andyw8:
There are legitimate cases why someone may want to edit code without the server running, so we want to avoid over-notifying. In this PR, we are changing the behaviour so that we output a notice in the logs, rather than showing an error dialog.

We should also update the README to explain how some features only available if the server is running, but since there is only currently one feature (Hover), it probably makes more sense to do that after Code Lens is added in https://github.com/Shopify/ruby-lsp-rails/pull/67.

Paired on with @vinistock

### DIFF
--- a/lib/ruby_lsp/ruby_lsp_rails/extension.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/extension.rb
@@ -13,7 +13,6 @@ module RubyLsp
 
       sig { override.void }
       def activate
-        # Must be the last statement in activate since it raises to display a notification for the user
         RubyLsp::Rails::RailsClient.instance.check_if_server_is_running!
       end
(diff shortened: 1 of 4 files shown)


lib/ruby_lsp/ruby_lsp_rails/extension.rb
```diff
@@ -13,7 +13,6 @@ class Extension < ::RubyLsp::Extension
 
       sig { override.void }
       def activate
-        # Must be the last statement in activate since it raises to display a notification for the user
```
andyw8, line 16: No longer relevant since we don't raise.

egiurleo (approved):
Nice! I think this will be a much better experience.

PR #89, opened 2023-06-22
Move to earlier in Rack middleware

andyw8:
Some applications may add custom middleware for processing requests, and the requests generated by ruby-lsp-rails may not match the expected structure, causing warnings or errors. (We were seeing this happen in Core).

By moving the middleware to earlier in the Rack middleware stack, we can prevent this.

There are a number of options for where we could have put this relevative to other middlewares - we went with `ActionDispatch::Static` as it is handled early in the request processing, and it is commonly used so unlikely to be removed.

We observed an additional benefit of a performance improvement since this is now handled before most of the Rails request processing.

Co-authored with @vinistock

### DIFF
--- a/lib/ruby_lsp_rails/railtie.rb
+++ b/lib/ruby_lsp_rails/railtie.rb
@@ -8,7 +8,7 @@ module RubyLsp
   module Rails
     class Railtie < ::Rails::Railtie
       initializer "ruby_lsp_rails.setup" do |app|
-        app.config.middleware.insert_after(ActionDispatch::ShowExceptions, RubyLsp::Rails::Middleware)
+        app.config.middleware.insert_before(ActionDispatch::Static, RubyLsp::Rails::Middleware)
 
         config.after_initialize do |_app|
           if defined?(::Rails::Server)


st0012:
I think the middleware is better not be injected earlier than `ActionDispatch::Executor`. It's responsible for running hooks registered to Rails' app executor. And the registered hooks including stuff like [reset execution context](https://github.com/rails/rails/blob/main/activesupport/lib/active_support/railtie.rb#L43). Given that our middleware does stuff like making database connection, it should be safer if we keep it inside those hooks.

st0012:
Will that avoid this issue? https://github.com/Shopify/ruby-dev-exp-issues/issues/898

rafaelfranca:
Yes. That would avoid that issue. I feel like we will be fighting a lot of application specific middleware configuration if we continue down to the path of making the LSP a middleware.

Yes, it is true that is the application has a middleware that sleeps for 30 seconds the LSP will be slower, so would be all the requests to the application. That is a problem the application should fix, not a concern that the LSP should try to avoid in detriment to having to know and workaround possible middleware stack configurations.

You don't necessarily need to implement the LSP as a Rails controller, just a Rack app mounted automatically, or manually in the routes would be good enough.

andyw8:
Changing to use a Rack app instead: https://github.com/Shopify/ruby-lsp-rails/pull/90

PR #90, opened 2023-06-23
Use a Rack app instead of Rails middleware

andyw8:
By using a Rack app instead of Rails Middleware, we don't need to worry about where to insert the middleware in the stack (https://github.com/Shopify/ruby-lsp-rails/pull/89).

Apps can opt of the app with `config.ruby_lsp_rails.server = false`, while still being able to take advantage of other features of the extension.

Tophatted against Code DB and Core (with the server disabled).

Co-authored with @vinistock

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,8 +10,22 @@ Ruby LSP Rails is a [Ruby LSP](https://github.com/Shopify/ruby-lsp) extension fo
 To install, add the following line to your application's Gemfile:
 
 ```ruby
+# Gemfile
 group :development do
   gem "ruby-lsp-rails"
+end
+```
+Some features rely on server introspection, and use a Rack server which is automatically mounted by using a Railtie.
+
+For applications with specialized routing requirements, such as custom sharding, this may not be compatible. It can
+be disabled with:
+
+```ruby
+# config/environments/development.rb
+Rails.application.configure do
+  # ...
+  config.ruby_lsp_rails.server = false
+  # ...
 end
 ```
 
@@ -41,7 +55,7 @@ cause the test runner to hang.
 
 This gem consists of two components that enable enhanced Rails functionality in the editor:
 
-1. A Rack middleware that automatically exposes APIs when Rails server is running
+1. A Rack app that automatically exposes APIs when Rails server is running
 1. A Ruby LSP extension that connects to the exposed APIs to fetch runtime information from the Rails server
 
 This is why the Rails server needs to be running for some features to work.
--- a/lib/ruby_lsp_rails/rack_app.rb
+++ b/lib/ruby_lsp_rails/rack_app.rb
@@ -3,32 +3,26 @@
 
 module RubyLsp
   module Rails
-    class Middleware
+    class RackApp
       extend T::Sig
 
       BASE_PATH = "/ruby_lsp_rails/"
-
-      sig { params(app: T.untyped).void }
-      def initialize(app)
-        @app = app
-      end
 
       sig { params(env: T::Hash[T.untyped, T.untyped]).returns(T::Array[T.untyped]) }
       def call(env)
         request = ActionDispatch::Request.new(env)
         path = request.path
-        return @app.call(env) unless path.start_with?(BASE_PATH)
 
         route, argument = path.delete_prefix(BASE_PATH).split("/")
 
         case route
+        when "activate"
+          [200, { "Content-Type" => "application/json" }, []]
         when "models"
           resolve_database_info_from_model(argument)
         else
-          [200, { "Content-Type" => "text/plain" }, []]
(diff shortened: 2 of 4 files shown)


lib/ruby_lsp_rails/rack_app.rb
```diff
@@ -3,32 +3,26 @@
...
       sig { params(env: T::Hash[T.untyped, T.untyped]).returns(T::Array[T.untyped]) }
       def call(env)
         request = ActionDispatch::Request.new(env)
         path = request.path
-        return @app.call(env) unless path.start_with?(BASE_PATH)
 
         route, argument = path.delete_prefix(BASE_PATH).split("/")
 
         case route
+        when "activate"
```
andyw8, line 19: Previously this was being handled by the `else` clause.

andyw8:
Moving back to draft as we need to update some things.

andyw8:
Internal issue for compatibility with Core: https://github.com/Shopify/ruby-dev-exp-issues/issues/978

lib/ruby_lsp_rails/rack_app.rb
```diff
@@ -3,32 +3,28 @@
...
       sig { params(env: T::Hash[T.untyped, T.untyped]).returns(T::Array[T.untyped]) }
       def call(env)
         request = ActionDispatch::Request.new(env)
         path = request.path
-        return @app.call(env) unless path.start_with?(BASE_PATH)
 
         route, argument = path.delete_prefix(BASE_PATH).split("/")
 
         case route
+        when nil # e.g. GET /ruby_lsp_rails
```
andyw8, line 19: Just to provide a simple way to verify the server is running.

README.md
```diff
@@ -10,10 +10,24 @@ Ruby LSP Rails is a [Ruby LSP](https://github.com/Shopify/ruby-lsp) extension fo
...
+Some features rely on server introspection, and use a Rack server which is automatically mounted by using a Railtie.
+
+For applications with specialized routing requirements, such as custom sharding, this may not be compatible. It can
+be disabled with:
+
+```ruby
+# config/environments.development.rb
+Rails.application.configure do
+  # ...
+  config.x.ruby_lsp_rails.server = false
```
andyw8, line 27: Using `config.x` as per the docs recommendation:

https://guides.rubyonrails.org/configuring.html

> The key difference between these two is that you should be using config.x if you are defining nested configuration (ex: config.x.nested.hi), and just config for single level configuration (ex: config.hello).

rafaelfranca: `x` is only for application config. Libraries should expose their own configs.

lib/ruby_lsp_rails/railtie.rb
```diff
@@ -2,15 +2,20 @@
 # frozen_string_literal: true
 
 require "rails/railtie"
-require "ruby_lsp_rails/middleware"
+require "ruby_lsp_rails/rack_app"
 
 module RubyLsp
   module Rails
     class Railtie < ::Rails::Railtie
-      initializer "ruby_lsp_rails.setup" do |app|
```
rafaelfranca, line 10: Let's add

```ruby
config.ruby_lsp_rails = ActiveSupport::OrderedOptions.new
config.ruby_lsp_rails.server = true
```

lib/ruby_lsp_rails/railtie.rb
```diff
@@ -2,15 +2,20 @@
...
+        config.after_initialize do |app|
+          unless app.config.x.ruby_lsp_rails.server == false
+            app.routes.prepend do
+              T.bind(self, ActionDispatch::Routing::Mapper)
+              mount(RackApp.new => RackApp::BASE_PATH)
+            end
+          end
 
-        config.after_initialize do |_app|
           if defined?(::Rails::Server)
```
rafaelfranca, line 19: Why do we need to do this?

andyw8: If we start the app with `bin/rails console` then the next line would fail with `uninitialized constant Rails::Server`.

lib/ruby_lsp_rails/rack_app.rb
vinistock: We check if the server is running by invoking `activate`. When is this branch called?

andyw8: Only when we hit `http://localhost:3000/ruby_lsp_rails`, e.g. in the browser to allow for a manual check. I can drop if you prefer not to have.

andyw8:
When tophatting I realised the config initialization needs to happen outside of the `ruby_lsp_rails.setup` block, updated in 3c5859adbe9feb794e69820e1702d968ba3ed236.

PR #95, opened 2023-06-28
Add link to docs

andyw8:


lib/ruby-lsp-rails.rb
```diff
@@ -8,6 +8,10 @@
 require "ruby_lsp_rails/railtie"
 
 module RubyLsp
+  # # Supported features
+  #
+  # - [Hover](rdoc-ref:RubyLsp::Rails::Hover)
+  # - [CodeLens](rdoc-ref:RubyLsp::Rails::CodeLens)
```
andyw8, line 14: @vinistock I added this as an equivalent to https://shopify.github.io/ruby-lsp/RubyLsp/Requests.html, so we have somewhere to link do directly, does it seem suitable?

vinistock: Yeah, great idea. I think this is the place to put it.

README.md
```diff
@@ -15,6 +15,7 @@ group :development do
 end
 ```
 
+
```
vinistock, line 18: ✂️

vinistock (approved):

PR #100, opened 2023-07-05
Fix Code Lens for test with empty string as name

andyw8:
The dashboard is showing some exceptions coming through for ruby-lsp-rails on Core. Although we are unable to see the exact code in use, I was able to reproduce with a test which has a blank string as the test name. If the Code Lens requests runs before the author has typed the test name then an exception occurs.

### DIFF
--- a/test/ruby_lsp_rails/code_lens_test.rb
+++ b/test/ruby_lsp_rails/code_lens_test.rb
@@ -8,6 +8,7 @@ module RubyLsp
     class CodeLensTest < ActiveSupport::TestCase
       setup do
         @message_queue = Thread::Queue.new
+        @store = RubyLsp::Store.new
       end
 
       def teardown
@@ -15,8 +16,7 @@ module RubyLsp
       end
 
       test "recognizes Rails Active Support test cases" do
-        store = RubyLsp::Store.new
-        store.set(uri: "file:///fake.rb", source: <<~RUBY, version: 1)
+        @store.set(uri: "file:///fake.rb", source: <<~RUBY, version: 1)
           class Test < ActiveSupport::TestCase
             test "an example" do
               # test body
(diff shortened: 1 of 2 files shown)


test/ruby_lsp_rails/code_lens_test.rb
```diff
@@ -38,17 +38,52 @@ class Test < ActiveSupport::TestCase
...
+          end
+        RUBY
+
+        response = RubyLsp::Executor.new(@store, @message_queue).execute({
+          method: "textDocument/codeLens",
+          params: { textDocument: { uri: "file:///fake.rb" }, position: { line: 0, character: 0 } },
+        }).response
+
+        # The 3 responses are for the test class, none for the test declaration.
+        assert_equal(3, response.size)
```
egiurleo, line 56: Excuse my ignorance on the testing API here -- is it possible to test the content of the responses to verify that they're coming from the test class? Otherwise it seems like the feature could break and the test could still potentially pass, but maybe there's no way to be more specific here.

egiurleo (commented):
Seems very reasonable to me but I have a question about the tests!

test/ruby_lsp_rails/code_lens_test.rb
andyw8: Good question - the other tests in this class check the response in more detail, but here the main intention was just to ensure that it doesn't raise an exception. There's a very slim chance that this single test could pass when it shouldn't, but in combination with the other tests I'm confident we'd detect the problem.

test/ruby_lsp_rails/code_lens_test.rb
```diff
@@ -38,17 +38,52 @@ class Test < ActiveSupport::TestCase
         assert_match("Debug", response[5].command.title)
       end
 
+      test "ignores unnamed tests (empty string)" do
```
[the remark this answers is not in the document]
andyw8: 👍 Added.

vinistock (approved):

PR #104, opened 2023-07-06
Support escaped multiline strings in test names

### LINKED ISSUE #103: Code lens doesn't handle multiline escaped strings
We're not handling this case, but it shouldn't be hard to support it.

```ruby
test "something" \
  "otherthing" do
end
```

vinistock:
Closes #103

I added support only for two lines. If we were to support any number of concatenations, we'd need to use a loop that keeps checking if the `right` side of the concatenation is another concatenation, which I think is not worth the complexity. Supporting one concatenation will be enough for the majority of cases.

### DIFF
--- a/lib/ruby_lsp/ruby_lsp_rails/code_lens.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/code_lens.rb
@@ -59,16 +59,26 @@ module RubyLsp
         return unless message_value == "test" && node.arguments.parts.any?
 
         first_argument = node.arguments.parts.first
-        return unless first_argument.is_a?(SyntaxTree::StringLiteral)
+
+        parts = case first_argument
+        when SyntaxTree::StringConcat
+          # We only support two lines of concatenation on test names
+          if first_argument.left.is_a?(SyntaxTree::StringLiteral) &&
+              first_argument.right.is_a?(SyntaxTree::StringLiteral)
+            [*first_argument.left.parts, *first_argument.right.parts]
+          end
+        when SyntaxTree::StringLiteral
+          first_argument.parts
+        end
 
         # The test name may be a blank string while the code is being typed
-        return if first_argument.parts.empty?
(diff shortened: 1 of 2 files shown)


lib/ruby_lsp/ruby_lsp_rails/code_lens.rb
```diff
@@ -59,16 +59,26 @@ def on_command(node)
         return unless message_value == "test" && node.arguments.parts.any?
 
         first_argument = node.arguments.parts.first
-        return unless first_argument.is_a?(SyntaxTree::StringLiteral)
+
+        parts = case first_argument
+        when SyntaxTree::StringConcat
+          # We only support two lines of concatenation on test names
```
andyw8, line 65: I wonder if it would be worth proposing something in Syntax Tree to make these easier to work with?

vinistock: I can't see an alternative structure. How could we make the tree simpler?

vinistock:
We wouldn't be able to show an error message or anything like that associated to code lens. Code lens are computed automatically for the entire file, so it could become quite annoying if we display a dialogue.

Do you feel like we should support deeper concatenations to avoid confusion? I'm hoping not a lot of people are writing >2 lines of test names 😂

andyw8 (approved):

egiurleo (approved):
@vinistock I think it makes sense to start here for now and see if anybody complains! I have to imagine there aren't a ton of people writing 3+ line test names out there 😅 Thanks for answering my question!

lib/ruby_lsp/ruby_lsp_rails/code_lens.rb
vinistock: I think it could potentially be in scope, but given that we're switching to YARP soon, let's move forward with this and re-evaluate later.

PR #106, opened 2023-07-12
Move listener registration to the extension activation

vinistock:
Related to https://github.com/Shopify/ruby-lsp/pull/813

We need to set a standard for registering feature enhancements always inside of `activate` or else disabling extensions becomes significantly more difficult.

This just moves our registrations there.

### DIFF
--- a/lib/ruby_lsp/ruby_lsp_rails/extension.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/extension.rb
@@ -14,6 +14,9 @@ module RubyLsp
 
       sig { override.void }
       def activate
+        ::RubyLsp::Requests::Hover.add_listener(RubyLsp::Rails::Hover)
+        ::RubyLsp::Requests::CodeLens.add_listener(RubyLsp::Rails::CodeLens)
+
         RubyLsp::Rails::RailsClient.instance.check_if_server_is_running!
       end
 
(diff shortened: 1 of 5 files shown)


lib/ruby_lsp/ruby_lsp_rails/extension.rb
```diff
@@ -14,6 +14,9 @@ class Extension < ::RubyLsp::Extension
 
       sig { override.void }
       def activate
+        ::RubyLsp::Requests::Hover.add_listener(RubyLsp::Rails::Hover)
```
Morriar, line 17: Since we only register classes and not instances, I wonder if this could be moved in the Listener itself:

```ruby
class RubyLsp::Rails::CodeLens < ::RubyLsp::Listener
  listen_to ::RubyLsp::Requests::CodeLens

  # ...
end
```

Morriar (approved):

lib/ruby_lsp/ruby_lsp_rails/extension.rb
vinistock: That's an interesting approach, but it would lead to the same problem we're trying to avoid in this PR.

In order to disable an extension and avoid registering the listeners defined by it, we need the registration to happen in `activate`.

Basically, if the server extension is disabled, we just don't invoke `activate`. If the listener classes register themselves, then that will happen automatically when we require the `Extension` file and we'll have no way of stopping it.

PR #131, opened 2023-08-23
Move DSL hover functionality from the Ruby LSP

vinistock:
In https://github.com/Shopify/ruby-lsp/pull/915, we are removing the Rails specific hover functionality in favour of the more generic one using the index. The Rails specific part should be in this extension.

This PR just moves the functionality in here. There's no difference other than how the tests are structured, since we don't have the same fixture mechanism in this extension.


andyw8 (approved):

lib/ruby_lsp/ruby_lsp_rails/support/rails_document_client.rb
```diff
@@ -0,0 +1,122 @@
...
+# frozen_string_literal: true
+
+require "net/http"
+
+module RubyLsp
+  module Rails
+    module Support
+      class RailsDocumentClient
+        RAILS_DOC_HOST = "https://api.rubyonrails.org"
+        SUPPORTED_RAILS_DOC_NAMESPACES = T.let(
```
Morriar, line 11: Adding a space to be consistent with the next constant:

```suggestion

        SUPPORTED_RAILS_DOC_NAMESPACES = T.let(
```

lib/ruby_lsp/ruby_lsp_rails/support/rails_document_client.rb
```diff
@@ -0,0 +1,122 @@
...
+            s.name == "railties"
+          end&.version&.to_s,
+          T.nilable(String),
+        )
+
+        class << self
+          extend T::Sig
+          sig do
+            params(name: String).returns(T::Array[String])
+          end
```
Morriar, line 36:
```suggestion
          extend T::Sig
          
          sig { params(name: String).returns(T::Array[String]) }
```

lib/ruby_lsp/ruby_lsp_rails/support/rails_document_client.rb
```diff
@@ -0,0 +1,122 @@
...
+
+          sig { returns(T.nilable(T::Hash[String, T::Array[T::Hash[Symbol, String]]])) }
+          private def build_search_index
+            return unless RAILTIES_VERSION
+
+            warn("Fetching Rails Documents...")
+
+            response = Net::HTTP.get_response(URI("#{RAILS_DOC_HOST}/v#{RAILTIES_VERSION}/js/search_index.js"))
+
+            if response.code == "302"
```
Morriar, line 73: In the docs they seem to favor comparing the class rather than the code:

```rb
case res
when Net::HTTPSuccess, Net::HTTPRedirection
  # OK
else
  res.value
end
```

vinistock:
Addressed the comments in https://github.com/Shopify/ruby-lsp-rails/pull/138.

lib/ruby_lsp/ruby_lsp_rails/support/rails_document_client.rb
andyw8: I think this is such a common thing that we should have a cop for it? I can create an issue if there isn't one already.

vinistock: I don't think there's an issue for it. Not sure how common it is, but definitely good to maintain standards.

andyw8: https://github.com/Shopify/rubocop-sorbet/issues/174

PR #132, opened 2023-08-23
Default uri path to localhost

vinistock:
For some reason, `URI("http://localhost:3000").path` returns an empty string. We need to default to using `localhost` in those cases or else requests to the Rails server fail.


vinistock:
It's very specific, but the `path` may be nil if there's an `opaque`. See [here](https://github.com/ruby/ruby/blob/5ec1fc52c1640c75899972ad99734ff22a90694a/lib/uri/generic.rb#L212).

PR #144, opened 2023-09-06
Support document symbol in ActiveSupport tests

st0012:
Unlike minitest, `ActiveSupport` tests are defined with `test` method, which
means they don't show up in the document symbols list provided by `ruby-lsp`. So this PR tries to add document symbol support to `ActiveSupport` tests.

However, to make the symbols have hierarchy, this extension also need to populate the class symbol (e.g. `FooTest`), which could be a duplication of `ruby-lsp`'s symbols:

Screenshot 2023-09-06 at 12 17 37

I think the solution should be to make `ruby-lsp` resolves `DocumentSymbol`'s responses more sophisticatedly, like merging nodes that have the same parent.

WDYT? @vinistock @andyw8


st0012:
@andyw8 could you be more specific about which point in that PR I should pay attention to?

PR #149, opened 2023-09-12
Upgrade to Ruby LSP v0.10.0

vinistock:
Upgrade our requirement to v0.10.0 and fix the breaking change (the rename from `@response` to `@_response` in listeners).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
         with:
           ruby-version: ${{ matrix.ruby }}
           bundler-cache: true
-          cache-version: 3
+          cache-version: 4
 
       - name: Check if documentation is up to date
         run: bundle exec rake ruby_lsp:check_docs


ruby-lsp-rails.gemspec
```diff
@@ -22,6 +22,6 @@ Gem::Specification.new do |spec|
   end
 
   spec.add_dependency("rails", ">= 6.0")
-  spec.add_dependency("ruby-lsp", ">= 0.9.1", "< 0.11.0")
+  spec.add_dependency("ruby-lsp", ">= 0.10.0", "< 0.11.0")
```
st0012, line 25: Not a blocker, but I think it's essentially `~> 0.10.0`?

st0012 (approved):

ruby-lsp-rails.gemspec
vinistock: It is, but for some reason dependabot always changes the requirement automatically to use `>=` and `<`.

PR #163, opened 2023-09-26
Remove Gemfile.lock when running CI on Ruby 3.1 Windows

vinistock:
It has proved impossible to get CI to work without removing the Gemfile.lock. Let's remove it only for Ruby 3.1 on Windows, so that our auto-merge continues to work.


andyw8:
I'm wondering if this will affect the bundle caching and possibly slow things down - we'll see once it's on `main`.

PR #224, opened 2024-01-09
Test against released Rails

andyw8:
As discussed in https://github.com/Shopify/ruby-lsp-rails/pull/220

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,6 +18,9 @@ jobs:
             experimental: true
           - gemfile: "gemfiles/Gemfile-rails-main"
             experimental: true
+        exclude:
+          - os: "windows-latest"
+            ruby: "3.0"
     runs-on: ${{ matrix.os }}
     continue-on-error: ${{ !!matrix.experimental }}
     name: Ruby ${{ matrix.ruby }} on ${{ matrix.os }} with ${{matrix.gemfile}}


test/dummy/db/migrate/20231019180159_create_posts.rb
```diff
@@ -1,4 +1,4 @@
-class CreatePosts < ActiveRecord::Migration[7.2]
+class CreatePosts < ActiveRecord::Migration[7.1]
```
andyw8, line 1: I ran a search and replace for `7.2` to `7.1`.

.github/workflows/ci.yml
```diff
@@ -9,13 +9,21 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
+        gemfile:
+          - Gemfile
+          - gemfiles/Gemfile-rails-main
```
andyw8, line 14: I followed the same naming as in `tapioca`.

andyw8:
(The branch protection rules will need updated once this is merged - we will need to bypass them for now).

gemfiles/Gemfile-rails-main
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+source "https://rubygems.org"
```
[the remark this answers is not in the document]
andyw8: Yeah, it's complains:

```
There was an error parsing `Gemfile-rails-main`: You cannot specify the same gem twice coming from different sources.
```

but we _could_ extract out the commonality, and call `eval_gemfile` from each?

(@Morriar do you have some approach for keeping these in sync on Tapioca?)

andyw8:
I've updated the branch protection rules.

PR #226, opened 2024-01-10
Remove Gemfile.lock for Ruby head

andyw8:
This will prevent failures such as:

```
nokogiri-1.16.0-x86_64-darwin requires ruby version < 3.4.dev, >= 3.0, which is
incompatible with the current version, 3.4.0.dev
```


paracycle (approved):

st0012 (approved):

.github/workflows/ci.yml
```diff
@@ -28,8 +28,8 @@ jobs:
       - uses: actions/checkout@v3
 
       - name: Remove Gemfile.lock
-        if: matrix.os == 'windows-latest' && matrix.ruby == '3.1'
-        run: rm Gemfile.lock
```
andyw8, line 32: This was invalid previously because it was removing the wrong lockfile in cases where the gemfile was `Gemfile-rails-main`. It seems the `gemspec` entry in the Gemfile wasn't playing well with a Gemfile contained in a subdirectory, so I moved it up to the root.

andyw8:
Ah, I should have used the `path` option.

https://github.com/Shopify/tapioca/blob/92857a3384e5755d735ecb2f6ed22451f95a7033/gemfiles/Gemfile-rails-main#L5C1-L5C19

PR #227, opened 2024-01-11
Test against ruby-lsp 0.13.3

st0012:
I noticed that due to https://github.com/Shopify/ruby-lsp/pull/1280, a test case here needs to be updated to pass with ruby-lsp [`0.13.3`](https://github.com/Shopify/ruby-lsp/releases/tag/v0.13.3).

The test didn't consider the constant's hover content because it wasn't displayed due to Sorbet being in the bundle. However, after the PR it's now displayed because the test isn't `typed: true`.

I decided not to fix it by adding `typed: true` because I think the tests here should reflect usages without Sorbet as it's what majority of the users would see.


andyw8:
Done in #228

PR #231, opened 2024-01-15
Support document symbol on ActiveSupport tests

st0012:
This feature allows users to navigate between AS tests with the `Go to Symbol` feature of VS Code.

It shares similar implementation with Code Lens around test case name extraction. And due to the complexity of the logic, I decided to extract it to a helper even though it's only shared between 2 components at the moment.

## Demo

document_symbol


lib/ruby_lsp/ruby_lsp_rails/document_symbol.rb
```diff
@@ -0,0 +1,43 @@
+# typed: strict
+# frozen_string_literal: true
+
+module RubyLsp
+  module Rails
+    # ![Document Symbol demo](../../document_symbol.gif)
+    #
+    # The [document symbol](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_documentSymbol)
```
vinistock, line 8: Let's not include the version and always link to the latest spec.
```suggestion
    # The [document symbol](https://microsoft.github.io/language-server-protocol/specification#textDocument_documentSymbol)
```

lib/ruby_lsp/ruby_lsp_rails/document_symbol.rb
```diff
@@ -0,0 +1,43 @@
...
+      include ActiveSupportTestCaseHelper
+
+      sig do
+        params(
+          response_builder: ResponseBuilders::DocumentSymbol,
+          dispatcher: Prism::Dispatcher,
+        ).void
+      end
+      def initialize(response_builder, dispatcher)
+        @_response = T.let(nil, NilClass)
```
vinistock, line 22: Do we need this variable?

vinistock (approved):
🚀

lib/ruby_lsp/ruby_lsp_rails/document_symbol.rb
st0012: Removed 👍

Issue #51, opened 2023-04-24
Add code lens to run or rollback migration


vinistock:
We can insert `Run` and `Rollback` code lenses on migrations for convenience.

PR #239, opened 2024-02-01
Add code lens for running migrations

gmcgibbon:
Closes https://github.com/Shopify/ruby-lsp-rails/issues/51

Adds code lens to run migrations to specific versions in the terminal. This is convenient because it allows developers to quickly rollback or fast-forward to specific schema versions with a click.

Screenshot 2024-01-31 at 9 09 05 PM
Screenshot 2024-01-31 at 9 10 45 PM

### DIFF
--- a/lib/ruby_lsp/ruby_lsp_rails/code_lens.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/code_lens.rb
@@ -10,6 +10,7 @@ module RubyLsp
     # - Run tests in the VS Terminal
     # - Run tests in the VS Code Test Explorer
     # - Debug tests
+    # - Run migrations in the VS Terminal
     #
     # The
     # [code lens](https://microsoft.github.io/language-server-protocol/specification#textDocument_codeLens)
@@ -31,6 +32,14 @@ module RubyLsp
     #   test "outputs goodbye" do # <- Will show code lenses above for running or debugging this test
     #     # ...
     #   end
+    # end
+    # ````
+    #
+    # # Example:
+    # ```ruby
+    # Run
+    # class AddFirstNameToUsers < ActiveRecord::Migration[7.1]
+    #   # ...
     # end
     # ````
     #
@@ -74,6 +83,7 @@ module RubyLsp
       sig { params(node: Prism::DefNode).void }
       def on_def_node_enter(node)
         method_name = node.name.to_s
+
         if method_name.start_with?("test_")
           line_number = node.location.start_line
           command = "#{test_command} #{@path}:#{line_number}"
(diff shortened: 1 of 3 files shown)


andyw8:
> Is there a different between migrating to a specific version vs rolling back to that version? For example, is migrating to the second to last version exactly the same as db:rollback?

I noticed this in the [docs](https://guides.rubyonrails.org/active_record_migrations.html):

> Neither of these rails commands do anything you could not do with db:migrate. They are there for convenience, since you do not need to explicitly specify the version to migrate to.

vinistock:
Thanks for sharing. Do people agree with using 3 code lenses?

```
Run all migrations | Migrate to this version | Rollback latest migration
```

I feel like usually people want to run `bin/rails db:migrate` and `bin/rails db:rollback`. Having those there will be handy, but it will also help communicate that `Migrate to this version` may actually be rolling back their database.

gmcgibbon:
Seems like my PRs are getting stale. I'll see if I can find some time to rebase and stuff the logic into different listeners.

gmcgibbon:
> Do people agree with using 3 code lenses?

Personally I would be supportive of this order: `Migrate to this version | Run all migrations | Rollback latest migration`. Idiomatically, if we're putting these lenses on all migrations, the developer would open the migration they want to go to. I would go as far as to say the other lenses should only be visible on the latest migration.

> Is there a difference between migrating using the latest migration version and simply running db:migrate to run all of them? I mean the end result of doing each action

Andy is correct, `VERSION=<latest>` and no args to run all are the same.

> Is there a different between migrating to a specific version vs rolling back to that version? For example, is migrating to the second to last version exactly the same as db:rollback?

Yes, I think so. Rollbacks can specify a step count though, so that's slightly different. The default step is one. It is just really convenient to use the `VERSION` for my use-case because Rails can figure out if I mean I want to go up or down.

lib/ruby_lsp/ruby_lsp_rails/code_lens.rb
```diff
@@ -31,13 +32,23 @@ module Rails
...
     # The code lenses will be displayed above the class and above each test method.
     class CodeLens < ::RubyLsp::Listener
       extend T::Sig
       extend T::Generic
 
       ResponseType = type_member { { fixed: T::Array[::RubyLsp::Interface::CodeLens] } }
-      BASE_COMMAND = "bin/rails test"
+      MIGRATE_COMMAND = "bin/rails db:migrate"
+      TEST_COMMAND = "bin/rails test"
+      BASE_COMMAND = TEST_COMMAND # TODO: Deprecate?
```
vinistock, line 51: We can definitely remove this. This is the only class that should be using this constant.

lib/ruby_lsp/ruby_lsp_rails/code_lens.rb
```diff
@@ -112,6 +131,34 @@ def on_class_node_leave(node)
...
+      sig { returns(T.nilable(String)) }
+      def migration_version
+        File.basename(T.must(@path)).split("_").first
+      end
+
+      sig { params(node: Prism::Node, name: String, command: String).void }
+      def add_migrate_code_lens(node, name:, command:)
+        return unless @path
+
+        arguments = [
```
vinistock, line 143: Does the command in the VS Code extension need the location information? If it's just using the command to run the migration, then we can remove the location information here and just return command.

vinistock (commented):
What if we go with a single code lens, but that says `Migrate to this version (2 versions behind latest)`?

I believe that would provide the full context and avoid needing multiple buttons.

vinistock (approved):
Awesome!

Issue #50, opened 2023-04-24
Add go to definition to connect controller actions to their routes and views


vinistock:
We can use go to definition to allow users to cmd click on top of a controller action and show the corresponding route and view as results. For example

```ruby
class PostsController < ApplicationController
  # CMD Clicking the index declaration could show
  # posts/index.html.erb
  # routes.rb:line_where_index_is_declared
  def index
  end
end
```

brunoprietog:
Yes, I imagine it's difficult, but at least I can dream. I just wanted to give that idea because those manual movements between views and controllers take a lot of time

Morriar:
Not a complete solution but any ERB code can be compiled into a Ruby code: 

```rb
require "erb"
erb = "<% x = 32 %><p><%= x %></p>"
ERB.new(erb).src # => "#coding:UTF-8\n_erbout = +'';  x = 32 ; _erbout.<< \"<p>\".freeze; _erbout.<<(( x ).to_s); _erbout.<< \"</p>\".freeze; _erbout"
```

vinistock:
Both of these were completed and shipped.

PR #241, opened 2024-02-03
Add code lens for mapping actions to routes

gmcgibbon:
Part of  https://github.com/Shopify/ruby-lsp-rails/issues/50

Adds code lens to find routes in the routes file based on a given controller action. You need to grep bin/rails routes for this information manually. If Rails already knows this info, we should make it more easily accessible.

Screenshot 2024-02-02 at 3 53 58 PM

### DIFF
--- a/lib/ruby_lsp/ruby_lsp_rails/code_lens.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/code_lens.rb
@@ -5,12 +5,22 @@ module RubyLsp
   module Rails
     # ![CodeLens demo](../../code_lens.gif)
     #
-    # This feature adds several CodeLens features for Rails applications using Active Support test cases:
+    # This feature adds Code Lens features for Rails applications.
+    #
+    # For Active Support test cases:
     #
     # - Run tests in the VS Terminal
     # - Run tests in the VS Code Test Explorer
     # - Debug tests
     # - Run migrations in the VS Terminal
+    #
+    # For Rails controllers:
+    #
+    # - See the path corresponding to an action
+    # - Click on the action's Code Lens to jump to its declaration in the routes.
+    #
+    # Note: This depends on a support for the `rubyLsp.openFile` command.
+    # For the VS Code extension this is built-in, but for other editors this may require some custom configuration.
     #
     # The
     # [code lens](https://microsoft.github.io/language-server-protocol/specification#textDocument_codeLens)
@@ -33,7 +43,7 @@ module RubyLsp
     #     # ...
     #   end
     # end
-    # ````
+    # ```
     #
     # # Example:
     # ```ruby
@@ -47,6 +57,20 @@ module RubyLsp
     #
     # Note: When using the Test Explorer view, if your code contains a statement to pause execution (e.g. `debugger`) it
     # will cause the test runner to hang.
+    #
+    # For the following code, assuming the routing contains `resources :users`, a Code Lens will be seen above each
+    # action.
+    #
+    # ```ruby
+    # class UsersController < ApplicationController
+    #   GET /users(.:format)
+    #   def index # <- Will show code lens above for the path
+    #   end
+    # end
+    # ```
+    #
+    # Note: Complex routing configurations may not be supported.
+    #
     class CodeLens
       extend T::Sig
       include Requests::Support::Common
@@ -54,16 +78,19 @@ module RubyLsp
 
       sig do
         params(
+          client: RunnerClient,
           response_builder:  ResponseBuilders::CollectionResponseBuilder[Interface::CodeLens],
           uri: URI::Generic,
           dispatcher: Prism::Dispatcher,
         ).void
       end
-      def initialize(response_builder, uri, dispatcher)
+      def initialize(client, response_builder, uri, dispatcher)
+        @client = client
         @response_builder = response_builder
         @path = T.let(uri.to_standardized_path, T.nilable(String))
         @group_id = T.let(1, Integer)
         @group_id_stack = T.let([], T::Array[Integer])
+        @constant_name_stack = T.let([], T::Array[[String, T.nilable(String)]])
 
         dispatcher.register(self, :on_call_node_enter, :on_class_node_enter, :on_def_node_enter, :on_class_node_leave)
       end
@@ -89,6 +116,10 @@ module RubyLsp
           command = "#{test_command} #{@path}:#{line_number}"
           add_test_code_lens(node, name: method_name, command: command, kind: :example)
         end
+
+        if controller?
+          add_route_code_lens_to_action(node)
+        end
       end
 
       sig { params(node: Prism::ClassNode).void }
@@ -107,17 +138,64 @@ module RubyLsp
           command = "#{migrate_command} VERSION=#{migration_version}"
           add_migrate_code_lens(node, name: class_name, command: command)
         end
+
+        # We need to use a stack because someone could define a nested class
+        # inside a controller. When we exit that nested class declaration, we are
+        # back in a controller context. This part is used in other places in the LSP
+        @constant_name_stack << [class_name, superclass_name]
       end
 
       sig { params(node: Prism::ClassNode).void }
       def on_class_node_leave(node)
         class_name = node.constant_path.slice
+
         if class_name.end_with?("Test")
           @group_id_stack.pop
         end
+
+        @constant_name_stack.pop
       end
 
       private
+
+      sig { returns(T.nilable(T::Boolean)) }
+      def controller?
+        class_name, superclass_name = T.must(@constant_name_stack.last)
+        class_name.end_with?("Controller") && superclass_name&.end_with?("Controller")
+      end
+
+      sig { params(node: Prism::DefNode).void }
+      def add_route_code_lens_to_action(node)
+        class_name, _ = T.must(@constant_name_stack.last)
+        route = @client.route(
+          controller: class_name,
+          action: node.name.to_s,
+        )
+
+        return unless route
+
+        path = route[:path]
+        verb = route[:verb]
+        source_location = route[:source_location]
+
+        arguments = [
+          source_location,
+          {
+            start_line: node.location.start_line - 1,
+            start_column: node.location.start_column,
+            end_line: node.location.end_line - 1,
+            end_column: node.location.end_column,
+          },
+        ]
+
+        @response_builder << create_code_lens(
+          node,
+          title: [verb, path].join(" "),
+          command_name: "rubyLsp.openFile",
+          arguments: arguments,
+          data: { type: "file" },
+        )
+      end
 
       sig { returns(String) }
       def test_command
--- a/test/ruby_lsp_rails/code_lens_test.rb
+++ b/test/ruby_lsp_rails/code_lens_test.rb
@@ -288,12 +288,29 @@ module RubyLsp
         assert_match("#{ruby} bin/rails db:migrate VERSION=123456", response[0].command.arguments[0])
       end
 
+      test "recognizes controller actions" do
+        response = generate_code_lens_for_source(<<~RUBY)
+          class UsersController < ApplicationController
+            def index
+            end
+          end
+        RUBY
+        path, line = response[0].command.arguments.first
+
+        assert_equal(1, response.size)
+        assert_match("GET /users(.:format)", response[0].command.title)
+        assert_equal("4", line)
+        assert_match("config/routes.rb", path)
+      end
+
       private
 
       attr_reader :ruby
 
       def generate_code_lens_for_source(source, file: "/fake.rb")
         with_server(source, URI(file)) do |server, uri|
+          sleep(0.1) while RubyLsp::Addon.addons.first.instance_variable_get(:@client).is_a?(NullClient)
+
           server.process_message(
             id: 1,
             method: "textDocument/codeLens",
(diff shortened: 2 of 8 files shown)


lib/ruby_lsp/ruby_lsp_rails/code_lens.rb
```diff
@@ -112,6 +119,50 @@ def on_class_node_leave(node)
...
+              start_column: node.location.start_column,
+              end_line: node.location.end_line - 1,
+              end_column: node.location.end_column,
+            },
+          ]
+
+          @_response << create_code_lens(
+            node,
+            title: [verb, path].join(" "),
+            command_name: "rubyLsp.openFile",
```
gmcgibbon, line 159: It seems kind of silly to me that there's no native support in editors for a file opening command (that I can see). Is there a way around this, or should I proceed with my PR that implements this over on the extension?

rickcaudill: I had a patch like this coming this week. I don't work for Shopify and have been really busy at work so I didn't have time to get the tests done yet, so I wasn't pushing it up.

The format I had for routes was:

            name: route.name,
            verb: route.verb,
            path: route.path.sub(/...format./, ""),
            controller: route.controller,
            action: route.action,
            source_location: "#{::Rails.application.root}/{route.source_location.sub(":", "#L")}",
            views: ActiveSupport::Inflector.safe_constantize((route.controller.to_s + "_controller").camelize).new.lookup_context.find_all("#{route.controller}/#{route.action}").map do |route|
                     { filename: route.identifier, format: route.format }
                   end,
          }

If you use vscode.open, you can do this for VS Code at least.  

If you want to do it all you could use just the uri for non vscode editors (but I haven't checked on how to do this yet).  

My lens looked like: 
Screenshot 2024-02-02 at 10 46 00 PM

rickcaudill: Can you please include the views as I have code action coming to open corresponding views?

rickcaudill: Also note, if you use #L#{line_number} you can jump to the right line.

rickcaudill: Either way I will be doing the views.  I am trying to replicate most of Rubymine for my team so I will probably do an add-on if needed.

rickcaudill: Thanks for the co-authorship :)

gmcgibbon:
Looks like this is failing on HEAD, but I don't know why yet.

lib/ruby_lsp/ruby_lsp_rails/code_lens.rb
```diff
@@ -42,12 +42,13 @@ class CodeLens < ::RubyLsp::Listener
...
 
-      sig { params(uri: URI::Generic, dispatcher: Prism::Dispatcher).void }
-      def initialize(uri, dispatcher)
+      sig { params(client: RailsClient, uri: URI::Generic, dispatcher: Prism::Dispatcher).void }
+      def initialize(client, uri, dispatcher)
         @_response = T.let([], ResponseType)
         @path = T.let(uri.to_standardized_path, T.nilable(String))
         @group_id = T.let(1, Integer)
         @group_id_stack = T.let([], T::Array[Integer])
+        @client = T.let(client, RailsClient)
```
andyw8, line 51:
```suggestion
        @client = client
```
(it's already typed by the sig)

lib/ruby_lsp/ruby_lsp_rails/code_lens.rb
```diff
@@ -112,6 +119,50 @@ def on_class_node_leave(node)
 
       private
 
+      sig { params(node: Prism::ClassNode).void }
+      def add_route_code_lenses_to_actions(node)
+        public_method_nodes = T.must(node.body).child_nodes.take_while do |node|
+          !node.is_a?(Prism::CallNode) || ![:protected, :private].include?(node.name)
+        end
+        public_method_nodes.grep(Prism::DefNode)
```
andyw8, line 127: Unused?

andyw8:
> This PR uses private API of Rails. I need to find a way to surface an endpoint in the Rails router to support this feature properly, if others agree it is useful.

When speaking with Jean and Aaron previously, they suggested that if necessary we can make modifications to Rails to expose an API for these kinds of things.

lib/ruby_lsp_rails/rack_app.rb
```diff
@@ -20,13 +20,37 @@ def call(env)
...
           not_found
         end
       end
 
       private
 
+      sig { params(requirements: T::Hash[Symbol, String]).returns(T::Array[T.untyped]) }
+      def resolve_route_info(requirements)
+        if requirements[:controller]
+          requirements[:controller] = T.unsafe(requirements[:controller]).underscore.delete_suffix("_controller")
```
andyw8, line 35:
```suggestion
          requirements[:controller] = requirements.fetch(:controller).underscore.delete_suffix("_controller")
```

lib/ruby_lsp/ruby_lsp_rails/code_lens.rb
```diff
@@ -112,6 +119,50 @@ def on_class_node_leave(node)
...
+
+      sig { params(node: Prism::DefNode, class_node: Prism::ClassNode).void }
+      def add_route_code_lense_to_action(node, class_node:)
+        route = @client.route(
+          controller: class_node.constant_path.slice,
+          action: node.name.to_s,
+        )
+
+        if route
+          path = route.dig(:path)
```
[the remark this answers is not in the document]
gmcgibbon: Originally, it was reading nested values. Will revert to using `[]`

andyw8:
:wave: @gmcgibbon 

I forgot about this when adding https://github.com/Shopify/ruby-lsp-rails/pull/331. There's a bit of commonality, so once that's merged we can hopefully rework this.

> This PR uses private API of Rails. 

Which aspect do you mean here?

lib/ruby_lsp/ruby_lsp_rails/code_lens.rb
andyw8: From looking at the [spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_codeLens), a Code Lens doesn't _need_ to have a command associated with it, so we could ship this with only the label, and deal with the command behaviour in a follow-up PR.

andyw8: @rickcaudill thanks for those suggestions. I think we can handle it separately, so I've created https://github.com/Shopify/ruby-lsp-rails/issues/384 to track.

andyw8:
Opened a PR for Rails: https://github.com/rails/rails/pull/51850

andyw8:
The Ruby 3.0 failures will gone with https://github.com/Shopify/ruby-lsp-rails/pull/391

andyw8:
> From looking at the [spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_codeLens), a Code Lens doesn't need to have a command associated with it, so we could ship this with only the label, and deal with the command behaviour in a follow-up PR.

Change of plan: https://github.com/Shopify/ruby-lsp/pull/2102

lib/ruby_lsp/ruby_lsp_rails/code_lens.rb
```diff
@@ -32,25 +40,41 @@ module Rails
...
     # The code lenses will be displayed above the class and above each test method.
     #
     # Note: When using the Test Explorer view, if your code contains a statement to pause execution (e.g. `debugger`) it
     # will cause the test runner to hang.
+    #
+    # For the following code, assuming the routing contains `resources :users`, a Code Lens will be seen above each
+    # action.
+    #
+    # ```ruby
+    # GET /users(.:format)
```
andyw8, line 54: Let's put this above the `def index` line?

test/ruby_lsp_rails/code_lens_test.rb
```diff
@@ -273,10 +273,25 @@ class Test < ActiveSupport::TestCase
...
+        assert_equal(1, response.size)
+        assert_match("GET /users(.:format)", response[0].command.title)
+        # assert_equal(["#{@client.root}/config/routes.rb", 3], response[0].command.arguments[0])
+      end
+
       private
 
       def generate_code_lens_for_source(source)
         with_server(source) do |server, uri|
+          sleep(0.1) while RubyLsp::Addon.addons.first.instance_variable_get(:@client).is_a?(NullClient)
```
andyw8, line 293: We now have https://github.com/Shopify/ruby-lsp/pull/2216 so we can simplify this a little:


```suggestion
          sleep(0.1) while RubyLsp::Addon.get("ruby-lsp-rails").is_a?(NullClient)
```

lib/ruby_lsp/ruby_lsp_rails/code_lens.rb
```diff
@@ -102,6 +132,48 @@ def on_class_node_leave(node)
 
       private
 
+      sig { params(node: Prism::ClassNode).void }
+      def add_route_code_lenses_to_actions(node)
+        public_method_nodes = T.must(node.body).child_nodes.take_while do |node|
```
andyw8, line 137: (not a blocker) I wonder if we should use the index to get the methods names 🤔

test/ruby_lsp_rails/code_lens_test.rb
```diff
@@ -273,10 +273,25 @@ class Test < ActiveSupport::TestCase
...
+        response = generate_code_lens_for_source(<<~RUBY)
+          class UsersController < ApplicationController
+            def index
+            end
+          end
+        RUBY
+
+        assert_equal(1, response.size)
+        assert_match("GET /users(.:format)", response[0].command.title)
+        # assert_equal(["#{@client.root}/config/routes.rb", 3], response[0].command.arguments[0])
```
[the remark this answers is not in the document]
gmcgibbon: Ah, right sorry. This was in your previous commits and I forgot to fix it. Will do now!

lib/ruby_lsp/ruby_lsp_rails/code_lens.rb
gmcgibbon: Done.

test/ruby_lsp_rails/code_lens_test.rb
gmcgibbon: Sure, done.

andyw8 (approved):
Looks great! Let's get @vinistock to review too though.

lib/ruby_lsp/ruby_lsp_rails/code_lens.rb
andyw8: Something like this: https://github.com/Shopify/ruby-lsp-rails/commit/debe164fcab9201adaf5467145c71534d12ba791

lib/ruby_lsp/ruby_lsp_rails/code_lens.rb
```diff
@@ -112,6 +119,50 @@ def on_class_node_leave(node)
...
+      end
+
+      sig { params(node: Prism::DefNode, class_node: Prism::ClassNode).void }
+      def add_route_code_lense_to_action(node, class_node:)
+        route = @client.route(
+          controller: class_node.constant_path.slice,
+          action: node.name.to_s,
+        )
+
+        if route
```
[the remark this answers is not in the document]
gmcgibbon: Done.

lib/ruby_lsp/ruby_lsp_rails/code_lens.rb
andyw8: Another option may be to do it server-side:

```
irb(main):001> UsersController.action_methods
=> #<Set: {"index", "archive", "create"}>
```

andyw8: A downside of the indexing approach is that currently we only reindex on save, so we wouldn't get the the immediate feedback when editing an action definition.

test/ruby_lsp_rails/code_lens_test.rb
```diff
@@ -288,12 +288,29 @@ def change
...
+        response = generate_code_lens_for_source(<<~RUBY)
+          class UsersController < ApplicationController
+            def index
+            end
+          end
+        RUBY
+        path, line = response[0].command.arguments.first
+
+        assert_equal(1, response.size)
+        assert_match("GET /users(.:format)", response[0].command.title)
```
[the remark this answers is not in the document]
gmcgibbon: `(.:format)` is an optional parameter. This way, the route can accept `/users` (no format), `/users.json`, and `/users.xml` and define behaviour based on the format provided. See https://guides.rubyonrails.org/routing.html#request-based-constraints, https://guides.rubyonrails.org/routing.html#defining-defaults

I left it as-is because otherwise the developer may not know it is there, and it matches what `bin/rails routes` outputs, but I'm happy to remove it if you think it isn't necessary.

lib/ruby_lsp/ruby_lsp_rails/code_lens.rb
gmcgibbon: - If a developer nests a class inside a controller class, does the leave event trigger at the right time so we don't decorate the wrong classes' methods?
- Why does the ivar need to be an array? It is either in a controller or it isn't, so we could use a single boolean right?

st0012 (approved):

test/ruby_lsp_rails/code_lens_test.rb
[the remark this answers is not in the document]
vinistock: Makes sense 👍

lib/ruby_lsp/ruby_lsp_rails/code_lens.rb
vinistock: Yes, I think we need to make this type of information available to all listeners. It's common for them to need it.

vinistock (approved):
Thank you! This is great

PR #243, opened 2024-02-03
Avoid inspecting schema info for abstract classes

gmcgibbon:
Hovering on model classes like `ApplicationRecord` in a Rails application results
in raising an `ActiveRecord::TableNotSpecifiedError,` which we should be
accounting for.

Before:
```
Started GET "/ruby_lsp_rails/models/ApplicationRecord" for ::1 at 2024-02-03 08:15:12 -0600

ActiveRecord::TableNotSpecified (ApplicationRecord has no table configured. Set one with ApplicationRecord.table_name=):
...
```

After:
```
Started GET "/ruby_lsp_rails/models/ApplicationRecord" for ::1 at 2024-02-03 08:17:35 -0600
```


st0012 (approved):
Thanks!

PR #250, opened 2024-02-09
WIP: Add Rails Runner client

andyw8:


.github/workflows/ci.yml
```diff
@@ -43,8 +43,8 @@ jobs:
           bundler-cache: true
           cache-version: 7
 
-      - name: Check if documentation is up to date
-        run: bundle exec rake ruby_lsp:check_docs
+      # - name: Check if documentation is up to date
+      #   run: bundle exec rake ruby_lsp:check_docs
```
andyw8, line 47: TODO: look into this

lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
```diff
@@ -0,0 +1,77 @@
...
+
+      sig { params(name: String).returns(T.nilable(T::Hash[Symbol, T.untyped])) }
+      def model(name)
+        make_request("model", name: name)
+      end
+
+      sig { void }
+      def shutdown
+        send_notification("shutdown")
+        sleep(0.01) while @wait_thread.alive?
```
vinistock, line 39: Just remembered that instead of sleeping we can directly ask Ruby to switch threads, which is a bit more elegant.
```suggestion
        Thread.pass while @wait_thread.alive?
```

lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
```diff
@@ -0,0 +1,77 @@
...
+
+      sig { params(request: T.untyped, params: T.untyped).returns(T.untyped) }
+      def make_request(request, params = nil)
+        send_message(request, params)
+        read_response
+      end
+
+      sig { params(request: T.untyped, params: T.untyped).void }
+      def send_message(request, params = nil)
+        json = { method: request, params: params }.compact.to_json
```
vinistock, line 53: I think compact might create another hash in between.
```suggestion
        message = { method: request }
        message[:params] = params if params
        json = message.to_json
```

lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
```diff
@@ -0,0 +1,80 @@
...
+      alias_method :send_notification, :send_message
+
+      sig { returns(T.nilable(T::Hash[Symbol, T.untyped])) }
+      def read_response
+        headers = @stdout.gets("\r\n\r\n")
+        raw_response = @stdout.read(T.must(headers)[/Content-Length: (\d+)/i, 1].to_i)
+
+        response = begin
+          JSON.parse(T.must(raw_response), symbolize_names: true)
+        rescue JSON::ParserError => e
```
andyw8, line 67: I think it's unlikely to fail here, but if it did it would result in a broken pipe error which could be difficult to diagnose.

andyw8:
This still needs a little clean-up but feedback is welcome.

lib/ruby_lsp/ruby_lsp_rails/server.rb
```diff
@@ -0,0 +1,70 @@
+# typed: strict
+# frozen_string_literal: true
+
+require "sorbet-runtime"
```
vinistock, line 4: Since this is a separate process, we need to add [this block](https://github.com/Shopify/ruby-lsp/blob/main/exe/ruby-lsp#L73-L86) to turn off runtime checks.

lib/ruby_lsp/ruby_lsp_rails/server_start.rb
vinistock, line 1: I would move this into the same file as `server` to be honest. No need to have separate files until this grows more complex.

test/ruby_lsp_rails/runner_client_test.rb
```diff
@@ -0,0 +1,51 @@
...
+        @client.shutdown
+
+        assert_predicate @client.instance_variable_get(:@stdin), :closed?
+        assert_predicate @client.instance_variable_get(:@stdout), :closed?
+        assert_predicate @client.instance_variable_get(:@stderr), :closed?
+        refute_predicate @client.instance_variable_get(:@wait_thread), :alive?
+      end
+
+      test "#model returns information for the requested model" do
+        # test/dummy/db/migrate/20230330202955_create_users.rb
```
vinistock, line 24: ✂️

lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
vinistock: IMO, this is a bit too defensive. We control the server and should never return invalid JSON from it.

test/ruby_lsp_rails/runner_client_test.rb
andyw8: I've updated the comment - I wanted to make it clear where the column names used in that test are coming.

.github/workflows/ci.yml
andyw8: t was because the docs task requires the `server`, which hangs the task. I've pushed a fix.

PR #253, opened 2024-02-12
Update Hover and CodeLens as per refactored responses in Ruby LSP

aryan-soni:
Based on Ruby LSP version 0.14.0, we need to update the `Hover` and `CodeLens` implementations in the Rails add-on.

These no longer extend the `Listener` class, and now use the response builder.

The associated tests have been updated, and these changes have been tested locally.

### DIFF
--- a/test/ruby_lsp_rails/hover_test.rb
+++ b/test/ruby_lsp_rails/hover_test.rb
@@ -44,16 +44,14 @@ module RubyLsp
           User
         RUBY
 
-        assert_equal(<<~CONTENT, response.contents.value)
+        assert_equal(<<~CONTENT.chomp, response.contents.value)
           ```ruby
           User
           ```
 
           **Definitions**: [fake.rb](file:///fake.rb#L1,1-2,4)
-
-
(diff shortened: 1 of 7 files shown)


test/ruby_lsp_rails/hover_test.rb
```diff
@@ -44,17 +44,15 @@ class User < ApplicationRecord
           User
         RUBY
 
-        assert_equal(<<~CONTENT, response.contents.value)
+        assert_equal(<<~CONTENT.chomp, response.contents.value)
```
andyw8, line 47: (nit) Why the change to using `chomp`? (I would have thought it better if the test shows the actual whitespace.)

vinistock: It's the heredoc literal that contains the extra line break. The actual result doesn't include it, which is a result of using `join` on the entries.

So showing the line break would require something like
```ruby
assert_equal(<<~CONTENT, "#{response.contents.value}\n")
```

PR #256, opened 2024-02-13
Switch to `rails runner` approach instead of Rack app

andyw8:
This PR switches Ruby LSP to the new pipe approach rather than a mounted Rack server. (the runner was added already in https://github.com/Shopify/ruby-lsp-rails/pull/250).

This has several advantages, such as not needing the Rails server to be running. Previous discussion can be found in the [prototype PR](https://github.com/Shopify/ruby-lsp-rails/pull/230).

If an app was configured with `config.ruby_lsp_rails.server`, this setting is no longer used and should be removed.

Note that there is currently no automatic reloading, so a change such as a database migration will not be visible until Ruby LSP is restarted.


lib/ruby_lsp_rails/railtie.rb
```diff
@@ -2,34 +2,23 @@
...
 
 require "rails"
-require "ruby_lsp_rails/rack_app"
 
 module RubyLsp
   module Rails
     class Railtie < ::Rails::Railtie
       config.ruby_lsp_rails = ActiveSupport::OrderedOptions.new
-      config.ruby_lsp_rails.server = true
+      # TODO: should we allow the runner to be enabled/disabled?
```
andyw8, line 10: TODO: remove all railtie

test/ruby_lsp_rails/hover_test.rb
```diff
@@ -2,13 +2,12 @@
 # frozen_string_literal: true
 
 require "test_helper"
+# require "ruby_lsp/ruby_lsp_rails/server"
```
vinistock, line 5: ✂️

vinistock (approved):

lib/ruby_lsp_rails/railtie.rb
```diff
@@ -1,38 +0,0 @@
-# typed: strict
-# frozen_string_literal: true
-
-require "rails"
-require "ruby_lsp_rails/rack_app"
-
-module RubyLsp
-  module Rails
-    class Railtie < ::Rails::Railtie
-      config.ruby_lsp_rails = ActiveSupport::OrderedOptions.new
```
st0012, line 10: Let's maybe keep the railtie so we can print deprecation warning for apps assigning `config.ruby_lsp_rails.server`?

I think upgrading `ruby-lsp-rails` and having the app blown up is not a good impression we want users to have.

st0012 (changes requested):

lib/ruby_lsp_rails/railtie.rb
andyw8: I was thinking that originally, but after discussing with @vinistock, probably very few apps are likely to be using this.

For Core, we can remove the setting when updating the gem.

We're also tagging this PR as a 'breaking change' for awareness.

st0012: > We're also tagging this PR as a 'breaking change' for awareness.

Yes, this and the version change will technically justify it, but they won't show up in the users' bump PR I think.

Given that:
1. This addon was [massively promoted](https://twitter.com/excid3/status/1753245795615125784) by Chris Oliver recently
2. Most Rails apps are private and won't show up in the GH search result

I don't want to underestimate the negative effect this may cause to our users, especially when it's actually not too costly to avoid.

vinistock: I highly doubt a lot of people are turning off the server, but if you feel strongly about this we can add a deprecation warning and then remove it later.

lib/ruby-lsp-rails.rb
```diff
@@ -1,11 +1,7 @@
 # typed: strict
 # frozen_string_literal: true
 
-require "sorbet-runtime"
```
andyw8, line 4: (`sorbet-runtime` is already required by `ruby-lsp`)

lib/ruby_lsp_rails/railtie.rb
```diff
@@ -0,0 +1,21 @@
...
+module RubyLsp
+  module Rails
+    class Railtie < ::Rails::Railtie
+      config.ruby_lsp_rails = ActiveSupport::OrderedOptions.new
+      config.ruby_lsp_rails.server = true
+
+      initializer "ruby_lsp_rails.setup" do |_app|
+        config.after_initialize do |app|
+          config.ruby_lsp_rails.server
+            raise "not needed!"
```
vinistock, line 15: Let's use a deprecation warning from Active Support instead. Otherwise, this will still break people's servers.

We can mention in the warning that the configuration can be removed.

lib/ruby_lsp_rails/railtie.rb
andyw8: Done (and verified against Code DB).

st0012: Thanks 👍 

I want to be super careful about this because I did already find 2 (though inactive) repos having this set just doing a [quick search](https://github.com/search?q=config.ruby_lsp_rails.server&type=code).

PR #258, opened 2024-02-15
Add code lens support for rails active support test cases using `minitest/spec`

beauraF:
Hello 👋

This is introducing code lens support for [minitest/spec](https://github.com/minitest/minitest?tab=readme-ov-file#specs-)

image

### DIFF
--- a/lib/ruby_lsp/ruby_lsp_rails/code_lens.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/code_lens.rb
@@ -57,7 +57,7 @@ module RubyLsp
       sig { params(node: Prism::CallNode).void }
       def on_call_node_enter(node)
         message_value = node.message
-        return unless message_value == "test"
+        return unless message_value == "test" || message_value == "it"
 
         arguments = node.arguments&.arguments
         return unless arguments&.any?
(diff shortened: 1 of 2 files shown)


beauraF:
Thanks @vinistock ! Sorry, I should have add more details ; it was planned, but haven't took the time to come back to it.

> Are the tests still expected to inherit from ActiveSupport::TestCase

All our tests still inherit from `ActiveSupport::TestCase`. We are using https://github.com/metaskills/minitest-spec-rails to ease the things and benefits from all the features of `minitest/spec`. But you can have a basic support by adding 
- `require "minitest/spec"` to `test/test_helper.rb` 
- `extend  extend Minitest::Spec::DSL` into `ActiveSupport::TestCase`

> Does describe work in this case? Do we need to handle it?

My proposal doesn't support `describe` but is able to launch a `it` inside a `describe`.
Supporting launching a group of tests from a `describe` would be a nice addition. But it's maybe a bit more work, and I wanted to start simply, without getting in the way of the initial philosophy of supporting vanilla `ActiveSupport::TestCase`. 
Now if that's something you would be ok with, I can work on that in a follow up PR. :)

PR #264, opened 2024-02-20
Handle `schema_dump_path` being unavailable in older Rails versions

### LINKED ISSUE #260: Rails 6.x compatibility broken due to `schema_dump_path`
From https://github.com/Shopify/ruby-lsp-rails/pull/250#discussion_r1494018448 the compatibility with Rails 6.x is broken.

The gemspec specify it is compatible with [ActiveRecord 6](https://github.com/Shopify/ruby-lsp-rails/blob/ba16bf3d9c4f150a1e9438c48165f2dcb926b192/ruby-lsp-rails.gemspec#L25) but the `schema_dump_path` is from ActiveRecord 7.0.

`schema_dump_path` is not present in ActiveRecord 6.1: https://github.com/rails/rails/blob/v6.1.7.6/activerecord/lib/active_record/tasks/database_tasks.rb

`schema_dump_path` is available from 7.0: https://github.com/rails/rails/blob/v7.0.0/activerecord/lib/active_record/tasks/database_tasks.rb#L448

Possible solutions:
- change the code to not rely on `schema_dump_path`
- change the gemspec to exclude Rails 6

andyw8:
Closes #260. For older Rails versions we will show the fields but won't provide a link to the schema file.

### DIFF
--- a/test/ruby_lsp_rails/runner_client_test.rb
+++ b/test/ruby_lsp_rails/runner_client_test.rb
@@ -16,7 +16,8 @@ module RubyLsp
         assert_predicate @client, :stopped?
       end
 
-      # This is an integration test which starts the server. For the more fine-grained tests, see `server_test.rb`.
+      # These are integration tests which start the server. For the more fine-grained tests, see `server_test.rb`.
+
       test "#model returns information for the requested model" do
         # These columns are from the schema in the dummy app: test/dummy/db/schema.rb
         columns = [


lib/ruby_lsp/ruby_lsp_rails/server.rb
```diff
@@ -26,7 +26,7 @@ class Server
 
       extend T::Sig
 
-      sig { params(model_name: String).returns(T.nilable(T::Hash[Symbol, T.untyped])) }
+      sig { params(model_name: String).returns(T::Hash[Symbol, T.untyped]) }
```
andyw8, line 29: I realised that we always return a hash.

test/ruby_lsp_rails/server_test.rb
```diff
@@ -0,0 +1,16 @@
+# typed: true
+# frozen_string_literal: true
+
+require "test_helper"
+require "ruby_lsp/ruby_lsp_rails/server"
+
+class ServerTest < ActiveSupport::TestCase
```
andyw8, line 7: @vinistock since the server is in a different process, this needs to be tested differently from the other aspects in test/ruby_lsp_rails/runner_client_test.rb. But I think it would make sense to move some of the others test to here too, and leave just model returns information for the requested model" as an integration test?

andyw8: Pushed a commit to show how that would look.

vinistock (approved):

test/ruby_lsp_rails/runner_client_test.rb
```diff
@@ -30,18 +31,6 @@ class RunnerClientTest < ActiveSupport::TestCase
...
-        assert_nil @client.model("Foo")
-      end
-
-      test "returns nil if class is not a model" do
-        assert_nil @client.model("Time")
-      end
-
-      test "returns nil if class is an abstract model" do
-        assert_nil @client.model("ApplicationRecord")
-      end
```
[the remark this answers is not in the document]
vinistock: Ah, right. It's covered in the new test. 👍

Issue #274, opened 2024-02-27
ruby-lsp at 100% CPU hanging in RubyLsp::Rails::RunnerClient#shutdown


tisba:
I cannot reliably reproduce this, but it happens ever now and then. When I just saw it happening again, I attached `rbspy` to record what's going on, here is the flame graph:

![CleanShot 2024-02-27 at 16 02 45](https://github.com/Shopify/ruby-lsp-rails/assets/75156/9dd66d8b-7734-4549-8643-8e5e9009f3f2)

It looks like it is spinning at `Thread.pass while @wait_thread.alive?` here https://github.com/Shopify/ruby-lsp-rails/blob/main/lib/ruby_lsp/ruby_lsp_rails/runner_client.rb#L67-L72.

In VSCode everything looks like it works normally. In the Ruby LSP output there is only one error, although I don't know if that's related to the issue:

```
/Users/sebastian/.gem/ruby/3.3.0/gems/ruby-lsp-0.14.3/lib/ruby_lsp/server.rb:112:in `push': queue closed (ClosedQueueError)
	from /Users/sebastian/.gem/ruby/3.3.0/gems/ruby-lsp-0.14.3/lib/ruby_lsp/server.rb:112:in `block in start'
	from /Users/sebastian/.gem/ruby/3.3.0/gems/language_server-protocol-3.17.0.3/lib/language_server/protocol/transport/io/reader.rb:20:in `read'
	from /Users/sebastian/.gem/ruby/3.3.0/gems/ruby-lsp-0.14.3/lib/ruby_lsp/server.rb:65:in `start'
	from /Users/sebastian/.gem/ruby/3.3.0/gems/sorbet-runtime-0.5.11268/lib/types/private/methods/_methods.rb:279:in `bind_call'
	from /Users/sebastian/.gem/ruby/3.3.0/gems/sorbet-runtime-0.5.11268/lib/types/private/methods/_methods.rb:279:in `block in _on_method_added'
	from /Users/sebastian/.gem/ruby/3.3.0/gems/ruby-lsp-0.14.3/exe/ruby-lsp:104:in `<top (required)>'
	from /Users/sebastian/.gem/ruby/3.3.0/bin/ruby-lsp:25:in `load'
	from /Users/sebastian/.gem/ruby/3.3.0/bin/ruby-lsp:25:in `<top (required)>'
[...]

PR #276, opened 2024-02-27
Set process group ID in server to terminate after parent

vinistock:
Closes https://github.com/Shopify/ruby-lsp-rails/issues/274

I think there are two issues with the way we shutdown the server. We can't have the infinite loop checking if the thread is alive, because something might go wrong with the server and then we'll never kill the child process. I removed that loop and switched to a single sleep to give some time for the server to shutdown.

Also, I started setting the process group ID in the server process. This is a similar trick used in [spring](https://github.com/rails/spring/blob/dee217996d2377e4f50bd61399d2431dbc6566d5/lib/spring/server.rb#L88), which makes it so that if the parent process is terminated, then so will the server.

I believe these two should help with the orphaned process issues we've been seeing.

### DIFF
--- a/lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
@@ -31,6 +31,7 @@ module RubyLsp
         # parent ends, the spring process ends as well. If this is not set, Spring will throw an error while trying to
         # set its own session ID
         begin
+          Process.setpgrp
           Process.setsid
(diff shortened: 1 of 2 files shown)


vinistock:
Actually, this doesn't work properly. We sometimes get an operation not permitted error. We'll need to think about alternatives.

vinistock:
@tisba my pleasure! Is the problem fixed on your end?

vinistock:
Sounds good. Thank you very much!

PR #280, opened 2024-03-03
Integrate document symbol for callbacks

aryan-soni:
Callbacks are ubiquitous across Rails, so it would be helpful to explicitly isolate them in the LSP via document symbols.

This approach accounts for callbacks associated with `ApplicationRecord` ([models](https://guides.rubyonrails.org/active_record_callbacks.html#available-callbacks)), `ApplicationController` ([controllers](https://api.rubyonrails.org/classes/AbstractController/Callbacks.html)), and `ApplicationJob` ([jobs](https://edgeapi.rubyonrails.org/classes/ActiveJob/Callbacks.html)).  

In addition to the various callback categories, there are 3 types of arguments to consider:
1. When the argument is a symbol (like `before_perform :foo_method`).
2. When the argument is a string (like `before_save "foo_method"`).
3. When the argument is a block (like   `before_action do ... end`).

So, this implementation looks like this:
1. Confirm the callback aligns with one of the callbacks offered by models/controllers/jobs.
2. Confirm the argument type is valid.
3. Create a document symbol for the callback-argument pairing.

As per this approach, something like `before_save "foo_method", "bar_method", on: :update` would yield two document symbols: `before_save(foo_method)` and `before_save(bar_method)`.

This approach directly ties to what is supported in Ruby LSP already for something like `attr_reader :foo`. 

See my thoughts below for additional context on the PR.

### DIFF
--- a/lib/ruby_lsp/ruby_lsp_rails/document_symbol.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/document_symbol.rb
@@ -12,6 +12,63 @@ module RubyLsp
       include Requests::Support::Common
       include ActiveSupportTestCaseHelper
 
+      MODEL_CALLBACKS = T.let(
+        [
+          "before_validation",
+          "after_validation",
+          "before_save",
+          "around_save",
+          "after_save",
+          "before_create",
+          "around_create",
+          "after_create",
+          "after_commit",
+          "after_rollback",
+          "before_update",
+          "around_update",
+          "after_update",
+          "before_destroy",
+          "around_destroy",
+          "after_destroy",
+          "after_initialize",
+          "after_find",
+          "after_touch",
+        ].freeze,
+        T::Array[String],
+      )
+
+      CONTROLLER_CALLBACKS = T.let(
+        [
+          "after_action",
+          "append_after_action",
+          "append_around_action",
+          "append_before_action",
+          "around_action",
+          "before_action",
+          "prepend_after_action",
+          "prepend_around_action",
+          "prepend_before_action",
+          "skip_after_action",
+          "skip_around_action",
+          "skip_before_action",
+        ].freeze,
+        T::Array[String],
+      )
+
+      JOB_CALLBACKS = T.let(
+        [
+          "after_enqueue",
+          "after_perform",
+          "around_enqueue",
+          "around_perform",
+          "before_enqueue",
+          "before_perform",
+        ].freeze,
+        T::Array[String],
+      )
+
+      CALLBACKS = T.let((MODEL_CALLBACKS + CONTROLLER_CALLBACKS + JOB_CALLBACKS).freeze, T::Array[String])
+
       sig do
         params(
           response_builder: ResponseBuilders::DocumentSymbol,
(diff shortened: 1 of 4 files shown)


lib/ruby_lsp/ruby_lsp_rails/support/active_support_test_case_helper.rb
```diff
@@ -11,8 +11,6 @@ def extract_test_case_name(node)
         message_value = node.message
         return unless message_value == "test" || message_value == "it"
 
-        return unless node.arguments
-
```
aryan-soni, line 15: This is not in the direct scope of this PR, but I've kept it to a separate commit; it is used by the `DocumentSymbol` class. The line below already accomplishments what this does.

lib/ruby_lsp/ruby_lsp_rails/document_symbol.rb
```diff
@@ -12,6 +12,61 @@ class DocumentSymbol
...
+          "after_enqueue",
+          "after_perform",
+          "around_enqueue",
+          "around_perform",
+          "before_enqueue",
+          "before_perform",
+        ].freeze,
+        T::Array[String],
+      )
+
```
aryan-soni, line 69: I've linked the source for each of these in the description, but the list is intended to be exhaustive.

Since callbacks remain relatively constant over time, this shouldn't be that suboptimal.

lib/ruby_lsp/ruby_lsp_rails/document_symbol.rb
```diff
@@ -28,13 +83,79 @@ def initialize(response_builder, dispatcher)
...
+        ).void
+      end
+      def append_document_symbol(name:, range:, selection_range:)
         @response_builder.last.children << RubyLsp::Interface::DocumentSymbol.new(
-          name: content,
-          kind: LanguageServer::Protocol::Constant::SymbolKind::METHOD,
-          selection_range: range_from_node(node),
-          range: range_from_node(node),
+          name: name,
+          kind: RubyLsp::Constant::SymbolKind::METHOD,
```
aryan-soni, line 156: This affects the icon of the symbol.

I chose the same approach as the base Ruby LSP. The options are listed in [this link](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#symbolKind) — I don't think it's optimal to use any other config.

vinistock (approved):
Nice!

lib/ruby_lsp/ruby_lsp_rails/document_symbol.rb
```diff
@@ -28,13 +83,79 @@ def initialize(response_builder, dispatcher)
...
+
+      private
+
+      sig { params(node: Prism::CallNode).void }
+      def extract_callbacks(node)
+        receiver = node.receiver
+        return if receiver && !receiver.is_a?(Prism::SelfNode)
+
+        message_value = node.message
+        callbacks = [MODEL_CALLBACKS, CONTROLLER_CALLBACKS, JOB_CALLBACKS]
```
andyw8, line 105: Could this be a constant?

andyw8:
Again not a blocker, but I notice it doesn't handle this syntax:

```ruby
before_create ->(user) { user.name = user.login.capitalize if user.name.blank? }
```
(from the [docs](https://guides.rubyonrails.org/active_record_callbacks.html))

andyw8:
I assume we've done some testing on Core with various syntaxes?

st0012 (commented):
Can we try splitting this into 2 classes to make things easier to understand in the future? Like `DslDocumentSymbol` and `TestDocumentSymbol`.
It should be doable and fairly simple under the new addon system. If not, we can also use the feedback from this attempt to improve it.

aryan-soni:
> Can we try splitting this into 2 classes to make things easier to understand in the future? Like `DslDocumentSymbol` and `TestDocumentSymbol`. It should be doable and fairly simple under the new addon system. If not, we can also use the feedback from this attempt to improve it.

I do agree that we have to create the abstraction at some point, but I'd rather scope it out as this PR already has a lot of points for discussion. After @andyw8's review, I noticed we should also add support for classes, like `before_save Foo`, alongside a couple other tangential paradigms.

I think it would be best to extract the classes out after I get to working on the other 2 dsl features.

aryan-soni:
> > I think it would be best to extract the classes out after I get to working on the other 2 dsl features.
> 
> Since other features are all dsl related, wouldn't this make the class separation more beneficial as the reviews will be easier?

My thinking is that it'll be easier to figure out what behaviour we can unify once we examine the behaviour of the other classes. In the dsl class, I don't think it'll be as simple as one method that applies to all of the features, so we can make decisions on the abstraction then.

I could separate things once I get to the next feature, instead of waiting until the very end. I agree that would make future reviews cleaner.

andyw8 (approved):

lib/ruby_lsp/ruby_lsp_rails/document_symbol.rb
```diff
@@ -12,6 +12,66 @@ class DocumentSymbol
...
+      MODEL_CALLBACKS = T.let(
+        [
+          "before_validation",
+          "after_validation",
+          "before_save",
+          "around_save",
+          "before_create",
+          "around_create",
+          "after_create",
+          "after_save",
```
andyw8, line 24: (nit) let's group with the other `save` hooks

lib/ruby_lsp/ruby_lsp_rails/document_symbol.rb
```diff
@@ -28,13 +88,115 @@ def initialize(response_builder, dispatcher)
...
+            selection_range: range_from_location(block.location),
+          )
+          return
+        end
 
+        arguments = node.arguments&.arguments
+        return unless arguments&.any?
+
+        arguments.each do |argument|
+          if argument.is_a?(Prism::SymbolNode)
```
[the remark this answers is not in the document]
aryan-soni: Good call out, that's certainly cleaner.

st0012 (approved):

Issue #257, opened 2024-02-14
Reloading approach for `rails runner`


andyw8:
We've now merged (but not yet released) https://github.com/Shopify/ruby-lsp-rails/pull/256 and it's working well.

We now need to consider how to deal with reloading, for example when running a migration which changes the schema.

Ideally we don't want to completely restart the runner, since that would be slow on large apps.

We're looking for a solution that could build upon the reloading mechanisms already present in Rails or Zeitwerk.

We could also make use of the LSP's ability to watch particular files for changes, which could trigger a notification to the `runner` process to reload particular files.

We'll need to do some research, but in the meantime @casperisfine we'd be interested to hear if you have any thoughts on this.

cc @vinistock

casperisfine:
> if you have any thoughts on this.

Right, so the classic autoloader used to try to be smart and remember the dependencies between files and only reload the ones that have changed or that depends on the ones that changed.

In theory with Zeitwerk you can still only unload the files you know have changed and it can reload them later on. The problem is that it open the doors to a lot of edge cases. e.g.

```foo.rb
class Foo
  def some_method
    1
  end
end
```

```bar.rb
class Bar
  FOO = Foo.new
end
```

If I edit `foo.rb`, you could only unload the `Foo` constant, but then `Bar::FOO` will continue to hold a reference to an instance of the old unloaded class. So you'd need to build a full dependency tree between the files or constant to try to optimize this without introducing weird edge cases.

I'd recommend first benchmarking unloading everything on changes on a reasonably sized application to see how well it would perform for the majority of users, and then start to optimize from there is really it doesn't seem viable.

For much larger applications like core, it may be acceptable to use heuristics that aren't always 100% correct.

Another thing to explore is whether everything need to be loaded upfront, maybe if you only load the constants you need you can get away with being totally lazy.

> for example when running a migration which changes the schema.

So for migrations specifically, you don't need to reload the code, just call `reset_column_information` on models: https://api.rubyonrails.org/classes/ActiveRecord/ModelSchema/ClassMethods.html#method-i-reset_column_information

vinistock:
This was completed in #282.

PR #282, opened 2024-03-06
Add reloading for schema changes

andyw8:
This change goes along with https://github.com/Shopify/ruby-lsp/pull/1464 in ruby-lsp.

Partly addresses https://github.com/Shopify/ruby-lsp-rails/issues/257

We register for changes to the schema files, and trigger a reload so as to get the latest column information. (Using the same mechanism that `reload!` in the Rails console uses).

### DIFF
--- a/lib/ruby_lsp/ruby_lsp_rails/addon.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/addon.rb
@@ -68,6 +68,15 @@ module RubyLsp
         DocumentSymbol.new(response_builder, dispatcher)
       end
 
+      sig { params(changes: T::Array[{ uri: String, type: Integer }]).void }
+      def workspace_did_change_watched_files(changes)
+        if changes.any? do |change|
+             change[:uri].end_with?("db/schema.rb") || change[:uri].end_with?("structure.sql")
+           end
+          @client.trigger_reload
+        end
+      end
+
       sig { override.returns(String) }
       def name
         "Ruby LSP Rails"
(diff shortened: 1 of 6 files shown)


test/ruby_lsp_rails/hover_test.rb
```diff
@@ -215,10 +215,6 @@ def hover_on_source(source, position)
         assert_nil(response.error)
         response.response
       end
-
-      def dummy_root
```
andyw8, line 219: Move into test_helper so that we can share.

lib/ruby_lsp/ruby_lsp_rails/addon.rb
```diff
@@ -68,6 +68,15 @@ def create_document_symbol_listener(response_builder, dispatcher)
         DocumentSymbol.new(response_builder, dispatcher)
       end
 
+      sig { params(changes: T::Array[{ uri: String, type: Integer }]).void }
+      def workspace_did_change_watched_files(changes)
+        if changes.any? do |change|
+             change[:uri].end_with?("db/schema.rb") || change[:uri].end_with?("structure.sql")
```
andyw8, line 74: Normally we would only have to support `structure.sql` but in `Shopify/shopify` we have additional structure files.

lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
```diff
@@ -75,6 +75,15 @@ def model(name)
         nil
       end
 
+      sig { void }
+      def trigger_reload
+        warn("Reloading Rails application")
+        send_notification("reload")
+      rescue IncompleteMessageError
+        warn("Ruby LSP Rails failed to get model information: #{@stderr.read}")
```
andyw8, line 83: Will correct this message.

PR #283, opened 2024-03-08
Implement document symbols for validations

aryan-soni:
Implements document symbol for the various types of Rails validations.

The remaining dsl feature is relationships. Based on my experience in this PR, it looks like the manner in which we handle each type of dsl feature doesn't generalize particularly cleanly. Consequently, after all the features are implemented, we can then reflect on the appropriate abstraction. Abstracting before comprehensive feature completion might cause us to create abstractions that we have to later rectify.

For now, these changes are scoped to focus on implementing validations exclusively.

### DIFF
--- a/lib/ruby_lsp/ruby_lsp_rails/document_symbol.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/document_symbol.rb
@@ -148,6 +148,8 @@ module RubyLsp
               selection_range: range_from_location(argument.location),
             )
           when Prism::CallNode
+            next unless argument.name == :new
+
             arg_receiver = argument.receiver
 
             name = arg_receiver.name if arg_receiver.is_a?(Prism::ConstantReadNode)
--- a/lib/ruby_lsp/ruby_lsp_rails/document_symbol.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/document_symbol.rb
@@ -93,25 +93,29 @@ module RubyLsp
           )
         end
 
-        extract_callbacks(node)
+        receiver = node.receiver
+        return if receiver && !receiver.is_a?(Prism::SelfNode)
+
+        message = node.message
+        case message
+        when *CALLBACKS, "validate"
+          handle_all_arg_types(node, T.must(message))
+        when "validates", "validates!", "validates_each"
+          handle_symbol_and_string_arg_types(node, T.must(message))
+        when "validates_with"
+          handle_class_arg_types(node, T.must(message))
+        end
       end
 
       private
 
-      sig { params(node: Prism::CallNode).void }
-      def extract_callbacks(node)
-        receiver = node.receiver
-        return if receiver && !receiver.is_a?(Prism::SelfNode)
-
-        message_value = node.message
-
-        return unless CALLBACKS.include?(message_value)
-
+      sig { params(node: Prism::CallNode, message: String).void }
+      def handle_all_arg_types(node, message)
         block = node.block
 
         if block
           append_document_symbol(
-            name: "#{message_value}(<anonymous>)",
+            name: "#{message}(<anonymous>)",
             range: range_from_location(node.location),
             selection_range: range_from_location(block.location),
           )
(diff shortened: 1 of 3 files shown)


lib/ruby_lsp/ruby_lsp_rails/document_symbol.rb
```diff
@@ -93,25 +93,29 @@ def on_call_node_enter(node)
...
-
-      private
-
-      sig { params(node: Prism::CallNode).void }
-      def extract_callbacks(node)
         receiver = node.receiver
         return if receiver && !receiver.is_a?(Prism::SelfNode)
 
-        message_value = node.message
+        message = node.message
```
andyw8, line 99:
```suggestion
        message = T.must(node.message)
```
then we can remove the `T.must`s below.

andyw8:
Although it began with https://github.com/Shopify/ruby-lsp-rails/pull/280, I'm now thinking more that we should have a visual distinction between macros and plain methods:
Screenshot 2024-03-11 at 3 43 49 PM

andyw8:
One slight oddity: if you have multiple validations for one field, such as:
```ruby
  validates :price, presence: true
  validates :price,
    numericality: { greater_than_or_equal_to: 0 },
```
Then you end up with multiple identical symbol entries.

`ProductVariant` in core has a lot of this. I can't think of a better approach, but just wanted to highlight it.

test/ruby_lsp_rails/document_symbol_test.rb
```diff
@@ -287,6 +287,68 @@ class FooJob < ApplicationJob
...
+        assert_equal(6, response[0].children.size)
+        assert_equal("validate(foo_arg)", response[0].children[0].name)
+        assert_equal("validate(bar_arg)", response[0].children[1].name)
+        assert_equal("validate(<anonymous>)", response[0].children[2].name)
+        assert_equal("validate(Foo::BazClass)", response[0].children[3].name)
+        assert_equal("validate(FooClass)", response[0].children[4].name)
+        assert_equal("validate(Foo::BarClass)", response[0].children[5].name)
+      end
+
+      test "correctly handles validates method with Prism::StringNode and Prism::SymbolNode argument types" do
```
andyw8, line 308:
```suggestion
      test "correctly handles validates method with string and symbol argument types" do
```
To avoid the being overly specific about the implementation.

test/ruby_lsp_rails/document_symbol_test.rb
```diff
@@ -287,6 +287,68 @@ class FooJob < ApplicationJob
...
+        RUBY
+
+        assert_equal(1, response.size)
+        assert_equal("FooModel", response[0].name)
+        assert_equal(2, response[0].children.size)
+        assert_equal("validates_each(foo_arg)", response[0].children[0].name)
+        assert_equal("validates_each(bar_arg)", response[0].children[1].name)
+      end
+
+      test "correctly handles validates_with method with Prism::ConstantReadNode and Prism::ConstantPathNode argument types" do
```
andyw8, line 338:
```suggestion
      test "correctly handles validates_with method with constants and namespaces constant argument types" do
```

aryan-soni:
> Although it began with #280, I'm now thinking more that we should have a visual distinction between macros and plain methods: Screenshot 2024-03-11 at 3 43 49 PM

I agree, nothing in the set of symbols offered seems like a good fit.

It is fairly easy to change down the road, so if something new is eventually released, it will be easy for us to change things.

lib/ruby_lsp/ruby_lsp_rails/document_symbol.rb
aryan-soni: Not sure about this one.

`node.message` is nilable. Once we enter the switch-case, we know for sure that it's not `nil`, but I don't think we can guarantee that before entering.

test/ruby_lsp_rails/document_symbol_test.rb
aryan-soni: This is also how the other tests for callbacks are structured, so I pushed a commit to fix this for them all.

lib/ruby_lsp/ruby_lsp_rails/document_symbol.rb
aryan-soni: Merged it without this, but we can fix this in the next PR based on your thoughts.

PR #284, opened 2024-03-08
Specifically handle the case where bin/rails is not found

st0012:
When the file is not found, it is likely that the LSP is not being run in the root of a Rails project. Having a more specific error message will help the user understand what is going on.

### DIFF
--- a/test/ruby_lsp_rails/runner_client_test.rb
+++ b/test/ruby_lsp_rails/runner_client_test.rb
@@ -60,9 +60,10 @@ module RubyLsp
         end
         File.chmod(0o755, "bin/rails")
 
+        # The error message is slightly different on Ubuntu, so we need to allow for that
         assert_output(
           "",
-          %r{Ruby LSP Rails failed to initialize server: bin/rails: line 1: foo:( command)? not found},
+          %r{Ruby LSP Rails failed to initialize server: bin/rails: (line )?1: foo:( command)? not found},
         ) do
           client = RunnerClient.create_client
 


andyw8:
I was working on this approach in ruby-lsp, so that if the app doesn't have `bin/rails` then we don't install the addon: https://github.com/Shopify/ruby-lsp/pull/1467.

lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
```diff
@@ -12,7 +12,15 @@ class << self
...
         sig { returns(RunnerClient) }
         def create_client
-          new
+          if File.exist?("bin/rails")
+            new
+          else
+            warn(<<~MSG)
+              Ruby LSP Rails failed to locate bin/rails in the current directory: #{Dir.pwd}"
+            MSG
+            warn("Server dependent features will not be available")
```
vinistock, line 21: We're trying to move away from `warn` and use `$stderr.puts` instead, so that we're not affected by the `VERBOSE` configuration. Can we please switch?

lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
```diff
@@ -12,7 +12,15 @@ class << self
 
         sig { returns(RunnerClient) }
         def create_client
-          new
+          if File.exist?("bin/rails")
```
vinistock, line 15: This will fail in workspace that are symlinks, since the addons don't yet have access to the workspace URI announced by the editor. There's nothing to do now, but I wanted to document this.

We need to create some sort of `State` object in the Ruby LSP that addons get initialized with, containing all of the initialization options, the message queue, etc.

vinistock (commented):
In regards to installing the addon or not, at the root of it is the question "is there value in having this addon even if it's not a Rails app?".

I think the answer might be yes because of the code lens support for Active Support style tests. But I'm not 100% sure to be honest. Thoughts?

lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
st0012: Should we make a dedicated PR to switch everything at once then? There are more than a dozen in the project atm.

st0012:
> I think the answer might be yes because of the code lens support for Active Support style tests. But I'm not 100% sure to be honest. Thoughts?

In addition to those, once we implemented something like go-to-definition to DSL symbols, like 

```rb
before_save :foo # cmd+click :foo jumps to foo method
```

Then it'll be useful in Rails engine development too.
And if we decided to support customising test code lens' executable name (e.g. `bin/test` instead of `bin/rails test`), it can directly benefit `rails/rails` and Rails engine devs too.

.rubocop.yml
```diff
@@ -35,6 +35,3 @@ Sorbet/StrictSigil:
   Exclude:
     - "**/*.rake"
     - "test/**/*.rb"
-
```
andyw8, line 38: Being removed in https://github.com/Shopify/ruby-lsp-rails/pull/285

andyw8:
Interesting, got a failure CI because on Ubuntu the error looks like:

`Ruby LSP Rails booting server\nRuby LSP Rails failed to initialize server: bin/rails: 1: foo: not found`
but on other platforms:
`Ruby LSP Rails booting server\nRuby LSP Rails failed to initialize server: bin/rails: line 1: foo: not found`

vinistock (approved):

PR #286, opened 2024-03-10
Add associations on Rails Models for Hover

rickcaudill:
This patch is to add associations to rails models: 

Screenshot 2024-03-10 at 12 27 54

### DIFF
--- a/lib/ruby_lsp/ruby_lsp_rails/hover.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/hover.rb
@@ -93,13 +93,14 @@ module RubyLsp
             )
           end
         end
-        @response_builder.push("---", category: :documentation)
-        @response_builder.push("---", category: :documentation)
 
-        # @response_builder.push(
-        # "[Schema](#{URI::Generic.build(scheme: "file", path: schema_file)})",
-        # category: :links,
-        # ) if schema_file
+        @response_builder.push(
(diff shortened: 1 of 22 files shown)


andyw8:
A few more suggestions after talking with the team:

* Rather than pushing line by line to `@response_builder`, let's build up the full string first, then we just need one call. It will probably help to extract a private method for this.
* Let's avoid the `nbsp` characters in the markdown. Even thought it may look in VS Code, this has to work in other editors too, so shouldn't assume too much about how it gets rendered.
* Using `heredoc`s rather the concatenating strings may improve readability.

andyw8:
:wave: @rickcaudill

At RailsConf next week, there is going to be Hack Days session for Ruby LSP Rails.

Would you be ok with other people contributing to this PR to help complete it?

rickcaudill:
> 👋 @rickcaudill
> 
> At RailsConf next week, there is going to be Hack Days session for Ruby LSP Rails.
> 
> Would you be ok with other people contributing to this PR to help complete it?

Actually I was going to get to this this weekend. If I don't happen to I would appreciate some help.  I have another one as well that I'm working on. The reason why its slow is because I'm in the middle of a big push at work and working with some additional clients right now. However I was planning to try to get this done this weekend. My team uses visual studio code as an IDE so these changes are quality of life changes for my team.

PR #293, opened 2024-03-15
Implement document symbol for associations

aryan-soni:
Implements the document symbol for various associations supported in Rails.

Any potential refactoring can be scoped out of this for now — this PR is focused on supporting the feature.

### DIFF
--- a/test/ruby_lsp_rails/document_symbol_test.rb
+++ b/test/ruby_lsp_rails/document_symbol_test.rb
@@ -349,6 +349,21 @@ module RubyLsp
         assert_equal("validates_with(Foo::BarClass)", response[0].children[1].name)
       end
 
+      test "correctly handles association callbacks with string and symbol argument types" do
+        response = generate_document_symbols_for_source(<<~RUBY)
+          class FooModel < ApplicationRecord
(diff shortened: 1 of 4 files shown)


test/ruby_lsp_rails/document_symbol_test.rb
```diff
@@ -349,6 +349,20 @@ class FooModel < ApplicationRecord
         assert_equal("validates_with(Foo::BarClass)", response[0].children[1].name)
       end
 
+      test "correctly handles association callbacks with string and symbol argument types" do
+        response = generate_document_symbols_for_source(<<~RUBY)
+          class FooModel < ApplicationRecord
+            belongs_to :foo, "baz"
```
andyw8, line 355: I'm sure Rails won't let you list multiple associations this way, so let's put them on separate lines.

andyw8:
Couple of observations when testing on Core:
- We won't have symbols for `accelerated_belongs_to`. Not a big deal, and we add if ever upstreamed to Rails.
- I discovered an existing issue in shopify-cops (see backlink).

andyw8 (approved):

PR #297, opened 2024-03-20
Add definitions for Rails callbacks

aryan-soni:
Implements definitions for callbacks in Rails, building on the approach leveraged for definitions inside `ruby-lsp`.

This PR also abstracts away the collection of Rails callbacks that we keep, since both the document symbol and definition features now rely on it. Moreover, generally, this PR introduces the definitions feature to the add-on, as it was not present previously.


andyw8:
`misc/definition.gif` is just copied from `ruby-lsp`, right? We should probably have one that's more focused on Rails features.

lib/ruby_lsp/ruby_lsp_rails/definition.rb
```diff
@@ -0,0 +1,94 @@
+# typed: strict
+# frozen_string_literal: true
+
+module RubyLsp
+  module Rails
+    # ![Definition demo](../../definition.gif)
```
vinistock, line 6: Let's change this documentation to mention only the feature contributed by `ruby-lsp-rails` (go to definition for callbacks).

lib/ruby_lsp/ruby_lsp_rails/definition.rb
```diff
@@ -0,0 +1,94 @@
...
+    # # Example
+    #
+    # ```ruby
+    # require "some_gem/file" # <- Request go to definition on this string will take you to the file
+    # Product.new # <- Request go to definition on this class name will take you to its declaration.
+    # ```
+    class Definition
+      extend T::Sig
+      include Requests::Support::Common
+      include ActiveSupportTestCaseHelper
```
vinistock, line 28: Do we need to include this?

test/ruby_lsp_rails/definition_test.rb
```diff
@@ -0,0 +1,125 @@
...
+module RubyLsp
+  module Rails
+    class DefinitionTest < ActiveSupport::TestCase
+      setup do
+        @message_queue = Thread::Queue.new
+
+        # Build the Rails documents index ahead of time
+        capture_io do
+          Support::RailsDocumentClient.send(:search_index)
+        end
```
vinistock, line 15: This isn't necessary for definition. It's only used in hover.

lib/ruby_lsp/ruby_lsp_rails/support/callbacks.rb
```diff
@@ -0,0 +1,67 @@
+# typed: strict
+# frozen_string_literal: true
+
+module RubyLsp
+  module Rails
+    module Support
+      module Callbacks
```
andyw8, line 7: Maybe `CallbacksList` to emphasize there's no behaviour in here?

lib/ruby_lsp/ruby_lsp_rails/document_symbol.rb
```diff
@@ -12,63 +12,7 @@ class DocumentSymbol
...
-          "around_enqueue",
-          "around_perform",
-          "before_enqueue",
-          "before_perform",
-        ].freeze,
-        T::Array[String],
-      )
-
-      CALLBACKS = T.let((MODEL_CALLBACKS + CONTROLLER_CALLBACKS + JOB_CALLBACKS).freeze, T::Array[String])
+      include Support::Callbacks
```
andyw8, line 15: Since we only use it one place, I think it's better to just reference it directly, e.g. `Support::Callbacks::CALLBACKS`. Makes it easier to see a glance where this is coming from.

(Also, we could rename `CALLBACKS` to `all` to avoid the repitition).

andyw8 (commented):
Looking good, just a few tweaks needed.

lib/ruby_lsp/ruby_lsp_rails/definition.rb
```diff
@@ -0,0 +1,87 @@
...
+    # definition of the symbol under the cursor.
+    #
+    # Currently supported targets:
+    # - Callbacks
+    #
+    # # Example
+    #
+    # ```ruby
+    # before_action :foo # <- Request go to definition on this symbol will take you to the method
+    # ```
```
andyw8, line 19: Let's add a note that this is currently only available if the method called method is defined in the same class?

vinistock (approved):
Nice 🚀

vinistock:
Let's add definition to [this list](https://github.com/Shopify/ruby-lsp-rails/blob/1f9b3e56c41ddb6a6a02ebe154eb4a5ef8d59d31/lib/ruby-lsp-rails.rb#L12).

andyw8 (approved):

PR #299, opened 2024-03-21
Use latest rdoc

andyw8:
For same reason as https://github.com/Shopify/ruby-lsp/pull/1810


andyw8:
Oh, @vinistock already updated in https://github.com/Shopify/ruby-lsp-rails/pull/298, not sure why the sha has changed?

andyw8:
Ah that would explain it.

PR #308, opened 2024-03-26
Set default CI timeout

andyw8:
To avoid jobs running continuously if there's a problem, like [this](https://github.com/Shopify/ruby-lsp/actions/runs/8440035436/job/23116030503).

(I think the default is 24h).


andyw8:
Actually meant this for `ruby-lsp`, but we may as well do it here too.

vinistock (approved):

PR #318, opened 2024-04-03
Only display code lens when the test framework is rails

st0012:
Currently, ruby-lsp-rails also sends code lens even when the test library is not rails. This creates problem when the user has both ruby-lsp-rspec and ruby-lsp-rails installed and the test framework is rspec, as the duplicated code lens will be cause none of them to display.

I also removed obsolete test setups that create/remove message queues manually. They're not used after the `with_server` is adopted.

### DIFF
--- a/test/ruby_lsp_rails/code_lens_test.rb
+++ b/test/ruby_lsp_rails/code_lens_test.rb
@@ -6,14 +6,6 @@ require "test_helper"
 module RubyLsp
   module Rails
     class CodeLensTest < ActiveSupport::TestCase
-      setup do
-        @message_queue = Thread::Queue.new
-      end
-
-      def teardown
-        T.must(@message_queue).close
-      end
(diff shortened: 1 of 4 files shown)


lib/ruby_lsp/ruby_lsp_rails/addon.rb
```diff
@@ -47,6 +47,8 @@ def deactivate
         ).void
       end
       def create_code_lens_listener(response_builder, uri, dispatcher)
+        return unless T.must(@global_state).test_library == "rails"
```
st0012, line 50: I decided to skip the entire listener because it's only used for test code lends atm.

st0012: I may be worth creating a separate listener for that feature now that we have conditions like this.

Issue #317, opened 2024-04-03
Review approach to testing


andyw8:
As reminded by https://github.com/Shopify/ruby-lsp-rails/pull/316, the current approach isn't very thorough since we are stubbing the runner responses.

PR #319, opened 2024-04-03
WIP: Revise testing approach

andyw8:
Closes https://github.com/Shopify/ruby-lsp-rails/issues/317

### DIFF
--- a/test/ruby_lsp_rails/hover_test.rb
+++ b/test/ruby_lsp_rails/hover_test.rb
@@ -14,20 +14,6 @@ module RubyLsp
       end
 
       test "hook returns model column information" do
-        expected_response = {
-          schema_file: "#{dummy_root}/db/schema.rb",
-          columns: [
-            ["id", "integer"],
-            ["first_name", "string"],
-            ["last_name", "string"],
-            ["age", "integer"],
-            ["created_at", "datetime"],
-            ["updated_at", "datetime"],
-          ],
-        }
-
-        RunnerClient.any_instance.stubs(model: expected_response)
-
         response = hover_on_source(<<~RUBY, { line: 3, character: 0 })
           class User < ApplicationRecord
           end
@@ -181,6 +167,10 @@ module RubyLsp
 
       def hover_on_source(source, position)
         with_server(source) do |server, uri|
+          while RubyLsp::Addon.addons.first.instance_variable_get(:@client).instance_of?(RubyLsp::Rails::NullClient)
+            Thread.pass
+          end
+
           server.process_message(
             id: 1,
             method: "textDocument/hover",


test/ruby_lsp_rails/hover_test.rb
```diff
@@ -181,6 +167,10 @@ class ActiveRecord::Base
 
       def hover_on_source(source, position)
         with_server(source) do |server, uri|
+          while RubyLsp::Addon.addons.first.instance_variable_get(:@client).instance_of?(RubyLsp::Rails::NullClient)
```
andyw8, line 170: We're doing the same in https://github.com/Shopify/ruby-lsp-rails/pull/303 but this should be cleaned up.

lib/ruby_lsp/ruby_lsp_rails/addon.rb
```diff
@@ -30,7 +30,9 @@ def activate(global_state, message_queue)
         @global_state = T.let(global_state, T.nilable(RubyLsp::GlobalState))
         $stderr.puts("Activating Ruby LSP Rails addon v#{VERSION}")
         # Start booting the real client in a background thread. Until this completes, the client will be a NullClient
-        Thread.new { @client = RunnerClient.create_client }
+        runner_thread = Thread.new { @client = RunnerClient.create_client }
+        # For tests, we want to wait for the client to be ready
+        runner_thread.join if ENV["RAILS_ENV"] == "test"
```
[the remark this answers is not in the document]
andyw8: (it does slow everything down though, so we'll probably need some way to selectively enable it...)

lib/ruby_lsp/ruby_lsp_rails/addon.rb
```diff
@@ -31,7 +31,7 @@ def activate(global_state, message_queue)
         @global_state = T.let(global_state, T.nilable(RubyLsp::GlobalState))
         $stderr.puts("Activating Ruby LSP Rails addon v#{VERSION}")
         # Start booting the real client in a background thread. Until this completes, the client will be a NullClient
-        Thread.new { @client = RunnerClient.create_client }.join
+        Thread.new { @client = RunnerClient.create_client }
```
andyw8, line 34: Being fixed in https://github.com/Shopify/ruby-lsp-rails/pull/356

test/ruby_lsp_rails/definition_test.rb
```diff
@@ -124,6 +124,8 @@ def baz; end
 
       def generate_definitions_for_source(source, position)
         with_server(source) do |server, uri|
+          sleep(0.1) while RubyLsp::Addon.addons.first.instance_variable_get(:@client).is_a?(NullClient)
```
andyw8, line 127: Being fixed in https://github.com/Shopify/ruby-lsp-rails/pull/356

test/ruby_lsp_rails/hover_test.rb
```diff
@@ -238,6 +151,8 @@ class ActiveRecord::Base
 
       def hover_on_source(source, position)
         with_server(source) do |server, uri|
+          sleep(0.1) while RubyLsp::Addon.addons.first.instance_variable_get(:@client).is_a?(NullClient)
```
andyw8, line 154: Same approach as we're using in `definition_test`, but we should really extract this to provide a better experience for addon developers.

test/ruby_lsp_rails/hover_test.rb
```diff
@@ -146,55 +100,14 @@ class CompositePrimaryKey < ApplicationRecord
...
-          **note**: string
+          **note**: text
 
           **created_at**: datetime
 
           **updated_at**: datetime
         CONTENT
       end
 
-      test "handles `db/structure.sql` instead of `db/schema.rb`" do
```
andyw8, line 157: With the new approach, these two tests would be awkward, and I don't think really provide much value, so I'm proposing we delete them. (If we did want to keep them, we'd should could them to `server_test.rb` and adapt them).

st0012: Yeah I think `server_test.rb` will be a better place for these tests. And since we do test against `schema_dump_path` being unsupported there, I think it's worth keeping these 2 tests there instead of deleting them.

test/ruby_lsp_rails/hover_test.rb
st0012: I agree that we should extract it into a helper. And I also think in that helper we should use `addons.find { |addon| addon.name == "Ruby LSP Rails" }` instead. I think `addons.first` will probably break if we someday add another addon to this project's Gemfile as a development dependency?

andyw8: Yeah, we can extract that part out too, e.g. `RubyLsp::Addon.get(...)`.

st0012: It does feel like something addons can use during testing 👍

andyw8: https://github.com/Shopify/ruby-lsp/pull/2216

PR #328, opened 2024-04-13
Prevent crash during ActiveRecord const check

andyw8:
I'm not exactly sure how it was triggered, but I had crash for this:

```
undefined method `<' for an instance of T::Private::Types::TypeAlias (NoMethodError)
```

I'll add a test if I'm able to reproduce.

### DIFF
--- a/lib/ruby_lsp/ruby_lsp_rails/server.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/server.rb
@@ -82,7 +82,7 @@ module RubyLsp
       sig { params(model_name: String).returns(T::Hash[Symbol, T.untyped]) }
       def resolve_database_info_from_model(model_name)
         const = ActiveSupport::Inflector.safe_constantize(model_name)
-        unless const && defined?(ActiveRecord) && const < ActiveRecord::Base && !const.abstract_class?
+        unless active_record_model?(const)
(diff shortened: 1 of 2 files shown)


lib/ruby_lsp/ruby_lsp_rails/server.rb
```diff
@@ -104,6 +104,17 @@ def resolve_database_info_from_model(model_name)
       rescue => e
         { error: e.full_message(highlight: false) }
       end
+
+      sig { params(const: T.untyped).returns(T::Boolean) }
+      def active_record_model?(const)
```
andyw8, line 109: Extracted out since this was getting a bit lengthy.

egiurleo (approved):

andyw8:
Updated based on @vinistock's suggestion, and I added a test.

st0012 (approved):
Solution looks great 👍 It'd be great to have some test cases around it if the setup is not too complicated.

st0012:
BTW, is `bug` the right label? I thought we use `bugfix`?

vinistock (approved):

st0012:
> is there something else you had in mind?

Oh sorry I must've clicked into just a specific commit. All good then 👍

PR #331, opened 2024-04-14
Add Definition support for routes

andyw8:
This PR adds support for the Definition request for Rails routes.

The implementation is based on @tenderlove's prototype in https://github.com/tenderlove/refreshing, and uses the behaviour added by @luanzeba and others in https://github.com/rails/rails/pull/47877.

Notes:

- For Sorbet projects using the Tapioca DSL generators for routes, the Definition Lookup will return two results - one for the RBI file, and one for the actual route.

- Unfortunately it doesn't work on Core, the source location always point to a `super` call in `routing_annotations.rb`. `bin/rails routes --expanded` has the same problem.

- There is some behaviour that might be a bug in VS Code:
If you double-click on a route name to select it, then choose Go To Definition, it works _only_ if the token has parenthesis on it, e.g. it works for `user_path(1)` but not `users_path`. But if you right click on `users_path` without selecting it, it works fine.

TODO:

- [x] confirm there is minimal performance impact for apps with a large number of routes
- [x] verify it doesn't break on older versions of Rails

# Testing

It's working well on Identity, (which doesn't use Sorbet) and Code DB (which does use Sorbet).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,10 +12,8 @@ jobs:
         gemfile:
           - Gemfile
           - gemfiles/Gemfile-rails-main
-        ruby: ["3.0", "3.1", "3.2", "3.3", "head"]
+        ruby: ["3.0", "3.1", "3.2", "3.3"]
         include:
-          - ruby: "head"
-            experimental: true
           - gemfile: "gemfiles/Gemfile-rails-main"
             experimental: true
         exclude:
--- a/lib/ruby_lsp/ruby_lsp_rails/addon.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/addon.rb
@@ -31,7 +31,7 @@ module RubyLsp
         @global_state = T.let(global_state, T.nilable(RubyLsp::GlobalState))
         $stderr.puts("Activating Ruby LSP Rails addon v#{VERSION}")
         # Start booting the real client in a background thread. Until this completes, the client will be a NullClient
-        Thread.new { @client = RunnerClient.create_client }
+        Thread.new { @client = RunnerClient.create_client }.join
       end
 
       sig { override.void }
@@ -84,7 +84,7 @@ module RubyLsp
       end
       def create_definition_listener(response_builder, uri, nesting, dispatcher)
         index = T.must(@global_state).index
-        Definition.new(response_builder, nesting, index, dispatcher)
+        Definition.new(@client, response_builder, nesting, index, dispatcher)
       end
 
       sig { params(changes: T::Array[{ uri: String, type: Integer }]).void }
--- a/lib/ruby_lsp/ruby_lsp_rails/definition.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/definition.rb
@@ -11,25 +11,35 @@ module RubyLsp
     #
     # Currently supported targets:
     # - Callbacks
+    # - Named routes (e.g. `users_path`)
     #
     # # Example
     #
     # ```ruby
     # before_action :foo # <- Go to definition on this symbol will jump to the method if it is defined in the same class
     # ```
+    #
+    # Notes for named routes:
+    # - It is available only in Rails 7.1 or newer.
+    # - Route may be defined across multiple files, e.g. using `draw`, rather than in `routes.rb`.
+    # - Routes won't be found if not defined for the Rails development environment.
+    # - If using `constraints`, the route can only be found if the constraints are met.
+    # - Changes to routes won't be picked up until the server is restarted.
     class Definition
       extend T::Sig
       include Requests::Support::Common
 
       sig do
         params(
+          client: RunnerClient,
           response_builder: ResponseBuilders::CollectionResponseBuilder[Interface::Location],
           nesting: T::Array[String],
           index: RubyIndexer::Index,
           dispatcher: Prism::Dispatcher,
         ).void
       end
-      def initialize(response_builder, nesting, index, dispatcher)
+      def initialize(client, response_builder, nesting, index, dispatcher)
+        @client = client
         @response_builder = response_builder
         @nesting = nesting
         @index = index
@@ -43,8 +53,19 @@ module RubyLsp
 
         message = node.message
 
-        return unless message && Support::Callbacks::ALL.include?(message)
+        return unless message
 
+        if Support::Callbacks::ALL.include?(message)
+          handle_callback(node)
+        elsif message.end_with?("_path") || message.end_with?("_url")
+          handle_route(node)
+        end
+      end
+
+      private
+
+      sig { params(node: Prism::CallNode).void }
+      def handle_callback(node)
         arguments = node.arguments&.arguments
         return unless arguments&.any?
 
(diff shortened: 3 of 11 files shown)


lib/ruby_lsp/ruby_lsp_rails/addon.rb
```diff
@@ -31,6 +31,7 @@ def activate(global_state, message_queue)
         $stderr.puts("Activating Ruby LSP Rails addon v#{VERSION}")
         # Start booting the real client in a background thread. Until this completes, the client will be a NullClient
         Thread.new { @client = RunnerClient.create_client }
+        @client = RunnerClient.create_client
```
andyw8, line 34: TODO: fix

andyw8:
Still a few things to finalize but this is ready for a first round of feedback.

lib/ruby_lsp/ruby_lsp_rails/server.rb
```diff
@@ -79,6 +81,36 @@ def execute(request, params)
 
       private
 
+      sig { params(name: String).returns(T::Hash[Symbol, T.untyped]) }
+      def route_location(name)
+        key = T.must(name.match(/^([a-zA-Z0-9_]+)(_path|_url)$/))[1]
```
andyw8, line 86: TODO: confirm what characters are permitted in route names.

lib/ruby_lsp/ruby_lsp_rails/definition.rb
```diff
@@ -9,27 +9,38 @@ module Rails
     # request](https://microsoft.github.io/language-server-protocol/specification#textDocument_definition) jumps to the
     # definition of the symbol under the cursor.
     #
+    # It is available only Rails 7.1 or newer.
```
andyw8, line 12: (move to next to routes part)

lib/ruby_lsp/ruby_lsp_rails/definition.rb
```diff
@@ -43,8 +54,19 @@ def on_call_node_enter(node)
 
         message = node.message
 
-        return unless message && Support::Callbacks::ALL.include?(message)
+        return unless message
+
+        if Support::Callbacks::ALL.include?(message)
+          handle_callback(node)
+        elsif message.match?(/^([a-zA-Z0-9_]+)(_path|_url)$/)
```
andyw8, line 61: (can simplifyy to just check ending)

lib/ruby_lsp/ruby_lsp_rails/definition.rb
```diff
@@ -62,7 +84,23 @@ def on_call_node_enter(node)
         end
       end
 
-      private
+      sig { params(node: T.untyped).void }
+      def handle_route(node)
+        result = @client.route_location(node.message)
+
+        location = T.must(result).fetch(:location)
```
andyw8, line 91: (Shouldn't need to do T.must here... look into)

lib/ruby_lsp/ruby_lsp_rails/server.rb
```diff
@@ -79,6 +81,37 @@ def execute(request, params)
...
+        end
+
+        key = T.must(name.match(/^([a-zA-Z0-9_]+)(_path|_url)$/))[1]
+
+        # A token could match the _path or _url pattern, but not be an actual route.
+        unless ::Rails.application.routes.named_routes.key?(key)
+          return { result: { location: nil } }
+        end
+
+        route = ::Rails.application.routes.named_routes.get(key)
```
andyw8, line 100: TODO: avoid calling `named_routes` twice.

andyw8: TODO: try avoid separate `key?` and `get?` calls.

lib/ruby_lsp/ruby_lsp_rails/server.rb
```diff
@@ -79,6 +81,33 @@ def execute(request, params)
 
       private
 
+      sig { params(name: String).returns(T::Hash[Symbol, T.untyped]) }
+      def route_location(name)
+        # Older versions of Rails don't support `route_source_locations`.
+        # We also check it hasn't been disabled.
+        unless ActionDispatch::Routing::Mapper.respond_to?(:route_source_locations) &&
+            ActionDispatch::Routing::Mapper.route_source_locations
```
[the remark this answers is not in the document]
andyw8: We can't rely purely on the version check because the app could have an initializer which sets `route_source_locations` to false.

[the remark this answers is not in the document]
andyw8: I think for this first iteration I'd like to leave the approach as-is, and we can re-visit.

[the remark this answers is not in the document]
st0012: If that's the case, then `ActionDispatch::Routing::Mapper.route_source_locations` needs to be in `server.rb` while `ActionDispatch::Routing::Mapper.respond_to?(:route_source_locations)` can be on the addon side. This is because only `server.rb` would be run with the application booted and initializers executed.

However, I also feel like it's not a user-facing option. Otherwise, it'd be accessible via `config` as well. From the [search result](https://github.com/search?q=repo%3Arails%2Frails%20route_source_locations&type=code), I feel that it's an internal option that's only controlled by Rails environment. If that's true, though, perhaps we don't even check its value as we'd always be running in dev environment anyway? 🤔

.github/workflows/ci.yml
```diff
@@ -12,10 +12,8 @@ jobs:
         gemfile:
           - Gemfile
           - gemfiles/Gemfile-rails-main
-        ruby: ["3.0", "3.1", "3.2", "3.3", "head"]
+        ruby: ["3.0", "3.1", "3.2", "3.3"]
```
andyw8, line 15: It seems that the behaviour added in https://github.com/rails/rails/pull/47877/files is not working properly with Ruby head. Since Rails itself doesn’t test against Ruby head, I think it’s reasonable if we don’t either, and wait until this is fixed upstream.

lib/ruby_lsp/ruby_lsp_rails/server.rb
```diff
@@ -79,6 +81,33 @@ def execute(request, params)
 
       private
 
+      sig { params(name: String).returns(T::Hash[Symbol, T.untyped]) }
+      def route_location(name)
+        # Older versions of Rails don't support `route_source_locations`.
+        # We also check it hasn't been disabled.
+        unless ActionDispatch::Routing::Mapper.respond_to?(:route_source_locations) &&
```
andyw8, line 88: TODO: move into a conditionally defined method once the PR to remove Sorbet is merged.

vinistock (approved):

lib/ruby_lsp/ruby_lsp_rails/addon.rb
```diff
@@ -31,7 +31,7 @@ def activate(global_state, message_queue)
         @global_state = T.let(global_state, T.nilable(RubyLsp::GlobalState))
         $stderr.puts("Activating Ruby LSP Rails addon v#{VERSION}")
         # Start booting the real client in a background thread. Until this completes, the client will be a NullClient
-        Thread.new { @client = RunnerClient.create_client }
+        Thread.new { @client = RunnerClient.create_client }.join
```
andyw8, line 34: (corrected in https://github.com/Shopify/ruby-lsp-rails/pull/356)

PR #338, opened 2024-04-16
Adjust Rails docs logging

andyw8:
* The trailing dots made it look like it's stuck doing something.
* Phrasing it as 'documents' felt a bit strange.


aryan-soni (approved):

vinistock (commented):
Nice. We've also received feedback a few times that people thought it was stuck because it says `Fetching docs`, but never says it finished doing it.

Should we add another print when it's done to avoid that confusion?

andyw8:
Updated with @vinistock's suggestion.

Issue #336, opened 2024-04-16
Failure for `sorbet-runtime` on new Rails app


andyw8:
I noticed when testing on a new Rails app:

```
2024-04-16 13:17:57.057 [info] (rails7) Ruby LSP Rails failed to initialize server: <internal:/opt/rubies/3.2.1/lib/ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:37:in `require': cannot load such file -- sorbet-runtime (LoadError)
```

I suspect this wasn't noticed previously because `sorbet-runtime` was already being required by something else.

vinistock:
This is strange. First, `sorbet-runtime` is required by `ruby-lsp` way before we even get close to activating any addons.

Second, this is saying that requiring `sorbet-runtime` has failed, which can only happen if the gem is not in the `$LOAD_PATH`, which should be impossible given that we have a dependency on `sorbet-runtime`.

andyw8:
Here's more of the stacktrace:

```
2024-04-16 13:17:57.057 [info] (rails7) Ruby LSP Rails failed to initialize server: <internal:/opt/rubies/3.2.1/lib/ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:37:in `require': cannot load such file -- sorbet-runtime (LoadError)
	from <internal:/opt/rubies/3.2.1/lib/ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:37:in `require'
	from /Users/andyw8/.gem/ruby/3.2.1/gems/bootsnap-1.18.3/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:17:in `require'
	from /Users/andyw8/.gem/ruby/3.2.1/gems/zeitwerk-2.6.13/lib/zeitwerk/kernel.rb:34:in `require'
	from /Users/andyw8/.gem/ruby/3.2.1/gems/ruby-lsp-rails-0.3.5/lib/ruby_lsp/ruby_lsp_rails/server.rb:4:in `<main>'
```

So it seems this is because in `server.rb` we're wrongly assuming that `sorbet-runtime` will be available in the main `Gemfile.lock`, when it may only be in `.ruby-lsp/Gemfile.lock`.

So either:
- We need to avoid using Sorbet completely within `server.rb`?
- We need to run `server.rb` but use with `BUNDLE_GEMFILE` set to `.ruby-lsp/Gemfile` ?

vinistock:
[...]

PR #340, opened 2024-04-16
WIP: Ensure sorbet-runtime available for runner

andyw8:
Alternative fix for https://github.com/Shopify/ruby-lsp-rails/issues/336. Still to test.


andyw8:
@vinistock this a little tricky to test because if I edit ".ruby-lsp/Gemfile" it will be overwritten when the LSP restarts... I seem to remember there was some workaround for that?

lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
```diff
@@ -48,7 +48,14 @@ def initialize
...
 
-        stdin, stdout, stderr, wait_thread = Open3.popen3(
+        env = {}
+        if ENV["BUNDLE_GEMFILE"]
+          env = { "BUNDLE_GEMFILE" => ENV["BUNDLE_GEMFILE"] }
+        end
+
+        # TODO: fix signature
+        stdin, stdout, stderr, wait_thread = T.unsafe(Open3).popen3(
+          env,
```
vinistock, line 58: I wonder if we can't simply pass `ENV` here and have it inherit the entire environment.

That would also make it easier to test, you could just add a bogus ENV value and verify that `popen3` gets invoked with it.

PR #341, opened 2024-04-17
Reload the runner if structure.sql changes

### LINKED ISSUE #326: Modifying `structure.sql` doesn't trigger an update event
Reloading for `schema.rb` and `structure.sql` was added in #282. The server however only listens for `*.rb` changes so it will never reload for `strucutre.sql` changes.

Culprit in the server code: https://github.com/Shopify/ruby-lsp/blob/729df3e909368ee09ba1315fa0a208dda52110ea/lib/ruby_lsp/server.rb#L206-L212

Demo video:

[Screencast_20240409_154242.webm](https://github.com/Shopify/ruby-lsp-rails/assets/14981592/dbef5e0f-1c10-44a3-ae9a-4ab8d734f32e)

Earlopain:
Closes #326

There already exists a test that modifications to structure.sql ends up reloading but without proper integration tests this behaviour isn't really testable.

To confirm this works:

* Get a rails app with structure.sql
* Make a change and save, look at the log output to validate it reloaded.
* Make some changes to a `*.rb` file to confirm that this new listener doesn't overwrite the main one.


andyw8:
Thank you! I've tested and it's working well.

lib/ruby_lsp/ruby_lsp_rails/addon.rb
```diff
@@ -96,6 +97,32 @@ def workspace_did_change_watched_files(changes)
         end
       end
 
+      sig { params(global_state: GlobalState, message_queue: Thread::Queue).void }
+      def register_additional_file_watchers(global_state:, message_queue:)
```
andyw8, line 101: Perhaps this could be `register_non_ruby_file_watchers`?

andyw8:
Let's also update `workspace_did_change_watched_files` to remove `structure.sql`? We can add a comment that it's only for Ruby files.

vinistock:
Just needs a rebase, but LGTM.

PR #342, opened 2024-04-17
Fix version logging

andyw8:
Prevent `VERSION` wrongly resolving to the one from ruby-lsp: https://github.com/Shopify/ruby-lsp/blob/main/lib/ruby-lsp.rb

### DIFF
--- a/lib/ruby_lsp/ruby_lsp_rails/addon.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/addon.rb
@@ -3,6 +3,7 @@
 
 require "ruby_lsp/addon"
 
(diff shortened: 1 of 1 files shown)


lib/ruby_lsp/ruby_lsp_rails/addon.rb
```diff
@@ -2,6 +2,7 @@
 # frozen_string_literal: true
 
 require "ruby_lsp/addon"
+require "ruby_lsp_rails/version"
```
st0012, line 5: Should this be `require_relative` like the below too?

st0012 (approved):

PR #344, opened 2024-04-17
Constrain sqlite3 gem for Rails

andyw8:
The v2.0.0 release of the [sqlite3 gem](https://rubygems.org/gems/sqlite3) today seems to be causing some issues with Rails ([example](https://github.com/Shopify/ruby-lsp-rails/actions/runs/8727205571/job/23944218243)). Let's temporarily constrain it until that is resolved.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,7 +7,7 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 gemspec
 
 gem "puma"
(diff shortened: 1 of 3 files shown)


andyw8:
It's also affecting the normal Gemfile so I've update it too (was likely only passing due to caching).

st0012 (approved):

PR #347, opened 2024-04-22
Exclude sqlite3 from Dependabot updates

andyw8:
v2 of the sqlite3 gem is not compatible with Rails 7.1.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -14,3 +14,5 @@ updates:
         update-types:
           - "minor"
           - "patch"
+        exclude-patterns:
+          - "sqlite3"


andyw8:
That's what we had but Dependabot still tries to bump it:
https://github.com/Shopify/ruby-lsp-rails/pull/346

It might be possible if we switch the [versioning strategy](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#versioning-strategy) but then it could be confusing if this repo behaves differently from others.

vinistock:
We don't auto-merge any major upgrades, which I think is fair so that someone has to understand what the major changes are. But we do want to keep dependencies up to date.

In my opinion, I wouldn't change the dependabot configuration. I'd just ignore the PR for sqlite until Rails supports it.

PR #350, opened 2024-04-24
Update tests for `ruby-lsp` changes

Earlopain:
Since https://github.com/Shopify/ruby-lsp/pull/1896 the testing framework for rails is dependant on the presence on `bin/rails`.

For https://github.com/Shopify/ruby-lsp-rails/pull/341 I need to bump the minimum required `ruby-lsp` but tests are failing due to the above PR, so I fix them here in a separate PR.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -10,7 +10,7 @@ PATH
   remote: .
   specs:
     ruby-lsp-rails (0.3.5)
-      ruby-lsp (>= 0.16.0, < 0.17.0)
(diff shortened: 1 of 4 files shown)


andyw8:
> Note: There are some green checks even though I'm pretty sure they should fail. I think tests doesn't run on windows right now because of how it is invoked in the workflow. It probably chokes on the `&&`.
> 
> https://github.com/Shopify/ruby-lsp-rails/actions/runs/8821159752

Thanks, logged as https://github.com/Shopify/ruby-lsp-rails/issues/351

Issue #351, opened 2024-04-24
Note: There are some green checks even though I'm pretty sure they should fail. I think tests doesn't run on windows right now because of how it is invoked in the workflow. It probably chokes on the `&&`.


andyw8:
Originally posted in https://github.com/Shopify/ruby-lsp-rails/pull/350

> Note: There are some green checks even though I'm pretty sure they should fail. I think tests doesn't run on windows right now because of how it is invoked in the workflow. It probably chokes on the `&&`.

https://github.com/Shopify/ruby-lsp-rails/actions/runs/8821159752

mohits:
> I remember this recent rails issue about the rubocop binstub not working on windows: https://github.com/rails/rails/issues/51618
> 
> Maybe `bundle binstubs rails --force` produces a meaningful result? Would probably need to be run on a windows machine, now that I think about it...

Hi, I opened https://github.com/rails/rails/issues/51618 and yes, `bundle binstubs rails --force` fixes it and everything runs fine after that.

Is there something you'd like me to try?

Thanks.

andyw8:
@mohits I am curious about this though:

If you create a new Rails app on Windows when using Powershell, what does the default `bin/rails` look like? And does running `bin/rails c` succeed?

andyw8:
> And finally, here's what Rails itself has to say about running rails under windows:

Thanks for finding that. The guides should probably be updated to reflect that this applies to any binstub provided by Rails, not just when starting the server.

Earlopain:
I'd say it does that already?. `bin/rails server` is used as an example since that's what the guide needs at that point but the part before makes it clear you need to do this for all comands under `bin`.

mohits:
Hi @andyw8 - just want to be sure that we're on the same page.

For the issue mentioned here, I do use `ruby bin\rails` for whatever I do, and the failure occurs when `rubocop` is invoked. However, this is permanently fixed for the project after you do this once:
```
bundle binstubs rubocop --force
```

No other change is needed anywhere else. As long as the conclusion you reached and the activity of https://github.com/Shopify/ruby-lsp-rails/pull/353 mean the same, it's fine.

A more detailed walkthrough is on [my page](https://notepad.onghu.com/2024/fix-rails-main-on-windows-rubocop-error/) about this issue.

andyw8:
@mohits I've replied on the [issue](https://github.com/rails/rails/issues/51618).

Issue #348, opened 2024-04-23
Can't use on windows (due to `setpgrp`)


krivospitsky:
I tried to use ruby-lsp in VSCode on windows and got this error:
```
ruby/3.2.0/gems/ruby-lsp-rails-0.3.5/lib/ruby_lsp/ruby_lsp_rails/addon.rb:33 run> terminated with exception (report_on_exception is true):
ruby/3.2.0/gems/ruby-lsp-rails-0.3.5/lib/ruby_lsp/ruby_lsp_rails/runner_client.rb:45:in `setpgrp': setpgrp() function is unimplemented on this machine (NotImplementedError)
```

andyw8:
Thank for the report. This was missed due to https://github.com/Shopify/ruby-lsp-rails/issues/351

andyw8:
@krivospitsky I don't have access to a Windows machine, so could you help me by checking something:

Can you clone this repo, then run `bin/rails c` to start the dummy app, and check if the console successfully starts. In CI it [seems to exit right away](https://github.com/Shopify/ruby-lsp-rails/actions/runs/8822162407/job/24219691837) (look under "Run Tests").

andyw8:
(the above is solved by https://github.com/Shopify/ruby-lsp-rails/pull/353)

andyw8:
@krivospitsky can you give [this branch](https://github.com/Shopify/ruby-lsp-rails/pull/352) a try by adding the following to your Gemfile:

```
gem "ruby-lsp-rails", github: "Shopify/ruby-lsp-rails", branch: "andyw8/fix-ci-setup-for-windows"
```

PR #352, opened 2024-04-24
Fix CI setup for Windows

andyw8:
Closes https://github.com/Shopify/ruby-lsp-rails/issues/351 but we still need to address https://github.com/Shopify/ruby-lsp-rails/issues/348.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,5 +54,8 @@ jobs:
       - name: Lint Ruby files
         run: bin/rubocop
 
+      - name: Lint Ruby files
+        run: bin/rubocop
+
       - name: Run tests
-        run: bin/rails db:setup && bin/rails db:migrate && bin/rails test
+        run: bundle exec rails db:setup && bundle exec rails db:migrate && bundle exec rails test
--- a/Gemfile
+++ b/Gemfile
@@ -20,3 +20,8 @@ gem "tapioca", "~> 0.13", require: false, platforms: :ruby
 gem "psych", "~> 5.1", require: false
 gem "rails"
 gem "webmock"
+
+platforms :mingw, :x64_mingw, :mswin, :jruby do
+  gem "tzinfo"
+  gem "tzinfo-data"
+end
--- a/lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
@@ -46,10 +46,13 @@ module RubyLsp
           Process.setsid
         rescue Errno::EPERM
           # If we can't set the session ID, continue
+        rescue NotImplementedError
+          # setpgrp() may be unimplemented on some platform
+          # https://github.com/Shopify/ruby-lsp-rails/issues/348
         end
 
         stdin, stdout, stderr, wait_thread = Bundler.with_original_env do
-          Open3.popen3("bin/rails", "runner", "#{__dir__}/server.rb", "start")
+          Open3.popen3("bundle", "exec", "rails", "runner", "#{__dir__}/server.rb", "start")
         end
 
         @stdin = T.let(stdin, IO)
--- a/test/ruby_lsp_rails/code_lens_test.rb
+++ b/test/ruby_lsp_rails/code_lens_test.rb
@@ -36,7 +36,7 @@ module RubyLsp
         # The last 3 are for the test declaration.
         assert_equal(6, response.size)
         assert_match("Run", response[3].command.title)
-        assert_equal("bin/rails test /fake.rb:2", response[3].command.arguments[2])
+        assert_match(%r{(ruby )?bin/rails test /fake\.rb:2}, response[3].command.arguments[2])
         assert_match("Run In Terminal", response[4].command.title)
         assert_match("Debug", response[5].command.title)
       end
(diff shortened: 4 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -24,7 +24,7 @@ jobs:
           - Gemfile: "gemfiles/Gemfile-rails-main" # needs Ruby 3.1
             ruby: "3.0"
     runs-on: ${{ matrix.os }}
-    timeout-minutes: 15
+    timeout-minutes: 20
```
andyw8, line 27: The CI is _really_ slow on Windows. Let's get it green first then look into.

test/ruby_lsp_rails/runner_client_test.rb
```diff
@@ -54,16 +54,10 @@ class RunnerClientTest < ActiveSupport::TestCase
...
 
       test "failing to spawn server creates a null client" do
-        FileUtils.mv("bin/rails", "bin/rails_backup")
-        File.open("bin/rails", "w") do |f|
-          f.write("foo")
-        end
-        File.chmod(0o755, "bin/rails")
-
-        # The error message is slightly different on Ubuntu, so we need to allow for that
+        FileUtils.mv("test/dummy/config/application.rb", "test/dummy/config/application.rb.bak")
```
andyw8, line 57: Picked this as an arbitrary file to remove to verify boot failure.

test/test_helper.rb
```diff
@@ -32,7 +32,11 @@ class TestCase
     include RubyLsp::TestHelper
 
     def dummy_root
-      File.expand_path("#{__dir__}/dummy")
+      result = File.expand_path("#{__dir__}/dummy")
+      unless result[0] == "/"
```
andyw8, line 36: The path returned on windows doesn't have the leading `/`.

andyw8:
Still seeing failures with the higher timeout... perhaps we are failing to shutdown the server properly on Windows, and it's using a ton of memory, leading to the slowness.

Gemfile
```diff
@@ -20,3 +20,4 @@ gem "tapioca", "~> 0.13", require: false, platforms: :ruby
 gem "psych", "~> 5.1", require: false
 gem "rails"
 gem "webmock"
+gem "tzinfo-data"
```
andyw8, line 23: Needed for Windows platforms. Trying to limit the platforms was causing some Bundler issues, so I've left it open.

vinistock: What issues?

lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
```diff
@@ -46,10 +46,15 @@ def initialize
...
         rescue Errno::EPERM
           # If we can't set the session ID, continue
+        rescue NotImplementedError
+          # setpgrp() may be unimplemented on some platform
+          # https://github.com/Shopify/ruby-lsp-rails/issues/348
         end
 
         stdin, stdout, stderr, wait_thread = Open3.popen3(
-          "bin/rails",
+          "bundle",
```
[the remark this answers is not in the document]
vinistock: This will bypass Spring and be extremely slow in Core. What exactly is the problem with using `bin/rails` on Windows?

test/test_helper.rb
vinistock: But why? The correct path on Windows wouldn't have that leading forward slash. Is it because of some URI being incorrect? If that's the case, then the solution is to use the Ruby LSP's URI builder `URI::Generic.from_path`, which already handles Windows URIs.

andyw8: Spoke to soon, still getting some windows failures.

andyw8:
Another possible approach is to prefix the binstub calls with `ruby`:
https://github.com/Shopify/ruby-lsp-rails/commit/bf9324b516dc422d180ad4d20b68dbe92c604337

I've verified this works in CI. It does kinda make sense that the binstub on its own wouldn't work on windows since the `#!/usr/bin/env ruby` is likely ignored... so does this mean there's no real way to write a cross-platform binstub?

test/ruby_lsp_rails/code_lens_test.rb
```diff
@@ -36,7 +36,7 @@ class Test < ActiveSupport::TestCase
         # The last 3 are for the test declaration.
         assert_equal(6, response.size)
         assert_match("Run", response[3].command.title)
-        assert_equal("bin/rails test /fake.rb:2", response[3].command.arguments[2])
+        assert_match(%r{(ruby )?bin/rails test /fake\.rb:2}, response[3].command.arguments[2])
```
andyw8, line 39: Due to https://github.com/Shopify/ruby-lsp-rails/pull/353

test/ruby_lsp_rails/runner_client_test.rb
```diff
@@ -4,6 +4,9 @@
 require "test_helper"
 require "ruby_lsp/ruby_lsp_rails/runner_client"
 
+# tests are hanging in CI. https://github.com/Shopify/ruby-lsp-rails/issues/348
```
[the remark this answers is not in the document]
vinistock: Do we still need to skip this test on Windows? If so, let's ship this, but please prioritize unskipping it so that we don't go long with skipped tests.

vinistock (approved):

PR #366, opened 2024-05-02
Set server pipes to binmode

vinistock:
There are a few Windows related fixes in this PR:

1. The STDIN and STDOUT pipes must be set to `binmode` to prevent conversion of carriage returns. Otherwise, the server communication is broken
2. There were some tests making assertions against URIs/paths that weren't valid on Windows
3. When splitting the file path on `:` to separate the line number, we need to account for the fact that Windows' paths begin with `C:/`, which was resulting in the wrong path and line number
4. `SIGTERM` cannot be used on Windows. Windows only responds to `KILL`, which we can also use on Linux/Mac to terminate a hanging server process

### DIFF
--- a/lib/ruby_lsp/ruby_lsp_rails/definition.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/definition.rb
@@ -88,7 +88,11 @@ module RubyLsp
         result = @client.route_location(T.must(node.message))
         return unless result
 
-        file_path, line = result.fetch(:location).split(":")
+        *file_parts, line = result.fetch(:location).split(":")
+
+        # On Windows, file paths will look something like `C:/path/to/file.rb:123`. Only the last colon is the line
+        # number and all other parts compose the file path
+        file_path = file_parts.join(":")
 
         @response_builder << Interface::Location.new(
           uri: URI::Generic.from_path(path: file_path).to_s,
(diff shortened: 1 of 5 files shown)


lib/ruby_lsp/ruby_lsp_rails/definition.rb
```diff
@@ -88,7 +88,8 @@ def handle_route(node)
         result = @client.route_location(T.must(node.message))
         return unless result
 
-        file_path, line = result.fetch(:location).split(":")
+        *file_parts, line = result.fetch(:location).split(":")
+        file_path = file_parts.join(":")
```
andyw8, line 92: Might be worth a comment to highlight that approach is needed because of Windows.

lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
```diff
@@ -80,7 +80,7 @@ def initialize
             if @wait_thread.alive?
               $stderr.puts("Ruby LSP Rails is force killing the server")
               sleep(0.5) # give the server a bit of time if we already issued a shutdown notification
-              Process.kill(T.must(Signal.list["TERM"]), @wait_thread.pid)
+              Process.kill(T.must(Signal.list["KILL"]), @wait_thread.pid)
```
andyw8, line 83: Let's mention that Windows doesn't support `KILL` in a comment?

andyw8 (approved):
Yay!

st0012 (approved):

lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
andyw8: (I mean `TERM`)

Issue #358, opened 2024-05-01
Provide jump-to-definition for associations


andyw8:
If you have some association such as:

`belong_to :user`
`has_many :products`

Then it would be good if you could use Go To Defintion on `:user` or `:products` to jump to to the corresponding model.

PR #367, opened 2024-05-03
Add a `has_many through` association to the dummy app

andyw8:
To help with manually testing https://github.com/Shopify/ruby-lsp-rails/issues/358


test/dummy/app/models/user.rb
```diff
@@ -2,7 +2,7 @@
 
 class User < ApplicationRecord
   before_create :foo, -> () {}
-  validates :name, presence: true
+  validates :first_name, presence: true
```
andyw8, line 5: (fixing invalid field)

PR #373, opened 2024-05-08
Add support for jump-to-def for associations

### LINKED ISSUE #358: Provide jump-to-definition for associations

maxveldink:
Closes #358 

Adds support for jump to definition for `belongs_to`, `has_one`, `has_many` and `has_and_belongs_to_many` associations on models.


maxveldink:
I have signed the CLA!

lib/ruby_lsp/ruby_lsp_rails/definition.rb
```diff
@@ -57,7 +57,9 @@ def on_call_node_enter(node)
 
         return unless message
 
-        if Support::Callbacks::ALL.include?(message)
+        if Support::Callbacks::ASSOCIATIONS.include?(message)
```
andyw8, line 60: Let's move this to `Support::Associations::ALL`?

andyw8:
:wave: Looking into the `on_class_node_enter` problem.

lib/ruby_lsp/ruby_lsp_rails/server.rb
```diff
@@ -110,6 +112,36 @@ def resolve_database_info_from_model(model_name)
...
+        const = ActiveSupport::Inflector.safe_constantize(params[:model_name])
+        unless active_record_model?(const)
+          return {
+            result: nil,
+          }
+        end
+
+        association_klass = case params[:association_type].intern
+        when :has_many
+          ActiveRecord::Associations::Builder::HasMany.build(const, params[:association_name].intern, nil, {}).klass
```
andyw8, line 125: Someone may have an invalid association specified, so we'll need handle that, e.g. by rescuing `NameError`.

lib/ruby_lsp/ruby_lsp_rails/server.rb
```diff
@@ -110,6 +112,42 @@ def resolve_database_info_from_model(model_name)
...
+            location: source_location.first + ":" + source_location.second.to_s,
+          },
+        }
+      rescue NameError => e
+        {
+          result: {
+            error: e.message,
+          },
+        }
+      end
```
maxveldink, line 149: I'm not sure the right behavior here; I opted to return the error message; but _I think_ would could return a nil result just like when the model name doesn't resolve properly. Thoughts?

andyw8: Yeah, I think `nil` is more appropriate here.

Thinks make me think though: Sometimes it's confusing for the user if they try something and the editor seems to do nothing. Perhaps there could be an `info` response, which prints something to the log in these kinds of situations.

maxveldink: Switched it to `nil` for now. I like the idea of having info responses print to the log. Should we add an issue to add that support?

lib/ruby_lsp/ruby_lsp_rails/definition.rb
```diff
@@ -85,24 +87,26 @@ def handle_callback(node)
         end
       end
 
+      sig { params(node: Prism::CallNode).void }
+      def handle_association(node)
+        association_name = T.cast(T.must(node.arguments).arguments.first, Prism::SymbolNode).unescaped
```
vinistock, line 92:
```suggestion
        first_argument = node.arguments&.arguments&.first
        return unless first_argument.is_a?(Prism::SymbolNode)
        
        
        association_name = first_argument.unescaped
```

lib/ruby_lsp/ruby_lsp_rails/definition.rb
```diff
@@ -85,24 +87,26 @@ def handle_callback(node)
...
+
+        result = @client.association_target_location(
+          model_name: @nesting.join("::"),
+          association_name: association_name,
+        )
+
+        return unless result&.key?(:location)
```
vinistock, line 99: If the result is not `nil`, then location is guaranteed to be present.
```suggestion
        return unless result
```

vinistock (commented):
Just a couple of comments. This looks great!

andyw8 (approved):

PR #374, opened 2024-05-09
Set up DB for tests as part of default Rake task

andyw8:
The first time someone cloned and runs the tests, they'll get a confusing error because the dB doesn't yet exist.

I originally considered checking if the DB exists first, but creating is very quick so I don't think that's necessary.

(for sqlite3 at least, `db:setup` is idempotent so there's no problem with it being re-run. We don't care about the data in the DB).

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -30,4 +30,4 @@ end
 
 RubyLsp::CheckDocs.new(FileList["#{__dir__}/lib/ruby_lsp/**/*.rb"], FileList["#{__dir__}/misc/**/*.gif"])
 
-task default: :test
+task default: [:"db:setup", :test]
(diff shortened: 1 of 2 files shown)


Rakefile
```diff
@@ -30,4 +30,4 @@ end
 
 RubyLsp::CheckDocs.new(FileList["#{__dir__}/lib/ruby_lsp/**/*.rb"], FileList["#{__dir__}/misc/**/*.gif"])
 
-task default: :test
+task default: [:"db:setup", :test]
```
andyw8, line 33: Previous we were running `migrate` but since `db:setup` calls `db:schema:load`, this isn't necessary.

Issue #378, opened 2024-05-13
Handle deferred route drawing


andyw8:
In https://github.com/rails/rails/pull/51614, @gmcgibbon has improved boot time by deferring route drawing. This will likely ship in the next release of Rails, so will require a small change for us.

For ruby-lsp-rails we could either:
* Defer loading until the first definition for a route (faster)
* Run `Rails.application.routes_reloader.execute_unless_loaded` (if defined) as part of `server.rb` to eagerly load the routes (slower).

andyw8:
Thanks @gmcgibbon. But for the LSP, we probably still want to eager load the routes so that the first request isn't delayed.

Looking at https://github.com/rails/rails/pull/52012

andyw8:
The Rails addon spawns a separate instance of the app, so anything we do there would have no impact on the Rails dev or test instances. It does mean that a developer using Ruby LSP will have to wait longer before _all_ the Rails LSP features are available, but that seems fine. 

We can however change [these lines ](https://github.com/Shopify/ruby-lsp-rails/blob/036cfd8d1f67c64142c1cc8571f40a20dd944b64/lib/ruby_lsp/ruby_lsp_rails/server.rb#L24-L25) to just `Rails.application.reload_routes!` to simplify things a little.

gmcgibbon:
Yeah that sounds fine to me. In the followup PR I introduced a method which might be helpful to not do it unless needed: `Rails.application.reload_routes_unless_loaded`, but it is private API.

PR #380, opened 2024-05-13
Eager load routes if necessary

andyw8:
The tests against Rails main are currently failing.

Closes https://github.com/Shopify/ruby-lsp-rails/issues/378


andyw8:
I've gone with the second option from https://github.com/Shopify/ruby-lsp-rails/issues/378 since doing it on the first request would introduce an awkward delay to the user experience.

egiurleo (approved):

PR #391, opened 2024-05-23
Drop Ruby 3.0 in CI

andyw8:
Ruby 3.0 is EOL as of 23 Apr 2024

We already dropped in ruby-lsp: https://github.com/Shopify/ruby-lsp/pull/2017

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,15 +12,10 @@ jobs:
         gemfile:
           - Gemfile
           - gemfiles/Gemfile-rails-main
-        ruby: ["3.0", "3.1", "3.2", "3.3"]
+        ruby: ["3.1", "3.2", "3.3"]
         include:
           - gemfile: "gemfiles/Gemfile-rails-main"
             experimental: true
-        exclude:
-          - os: "windows-latest"
-            ruby: "3.0"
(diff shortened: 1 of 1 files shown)


ruby-lsp-rails.gemspec
```diff
@@ -17,7 +17,7 @@ Gem::Specification.new do |spec|
   spec.metadata["source_code_uri"] = spec.homepage
   spec.metadata["changelog_uri"] = "#{spec.homepage}/releases"
 
-  spec.required_ruby_version = ">= 3.0.0"
+  spec.required_ruby_version = ">= 3.1.0"
```
vinistock, line 20: We didn't drop support for Ruby 3.0 in the Ruby LSP, we just removed it from CI because `rbi` dropped support for it and bundle install fails.

I know 3.0 is already EOL, but can we please wait a little longer and avoid another situation like Ruby 2.7?

There are no features we want to use that depend on Ruby > 3.0 anyway.

vinistock (changes requested):

ruby-lsp-rails.gemspec
andyw8: Restored.

vinistock (approved):

PR #414, opened 2024-07-23
Rescue `IOError` during shutdown

andyw8:
If the server has died (e.g. after running for a long time) then an `IOError` may occur while trying to shut it down.

### DIFF
--- a/lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
@@ -84,9 +84,7 @@ module RubyLsp
             if @wait_thread.alive?
               $stderr.puts("Ruby LSP Rails is force killing the server")
               sleep(0.5) # give the server a bit of time if we already issued a shutdown notification
-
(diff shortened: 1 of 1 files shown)


lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
```diff
@@ -149,6 +149,8 @@ def shutdown
         send_message("shutdown")
         sleep(0.5) # give the server a bit of time to shutdown
         [@stdin, @stdout, @stderr].each(&:close)
+      rescue IOError
+        # The server connection may have died
```
vinistock, line 153: I wonder if we should still issue a kill signal to guarantee we don't leave orphan processes. Could there be a scenario where we get an `IOError`, but the child process is still alive?

Maybe we can do this inside the rescue?
```ruby
Process.kill(T.must(Signal.list["KILL"]), @wait_thread.pid)
```

We can also ask the reporter to modify their local installation of `ruby-lsp-rails` so that we can verify that it works.

andyw8:
(waiting for feedback from reporter before merging)

PR #423, opened 2024-07-24
Add association augmentations

vinistock:
This is another example of how we can use indexing augmentations https://github.com/Shopify/ruby-lsp/pull/2358.

This PR starts handling association DSLs such as has_one and belongs_to, creating the accessor methods that are generated dynamically when invoking them.


st0012 (approved):
Do we plan to add more methods like `build_user`, `create_user`....etc. in the future too?

PR #426, opened 2024-07-26
Fix `NullClient#send_notification`

### LINKED ISSUE #425: Calling `send_notification` too early causes a `NoMethodError`
# Error

```
Error processing workspace/didChangeWatchedFiles: /Users/alex/src/github.com/Shopify/ruby-lsp-rails/lib/ruby_lsp/ruby_lsp_rails/runner_client.rb:189:in `send_message': undefined method `write' for nil (NoMethodError)

        @stdin.write("Content-Length: #{json.length}\r\n\r\n", json)
              ^^^^^^
	from /Users/alex/.gem/ruby/3.3.3/gems/sorbet-runtime-0.5.11465/lib/types/private/methods/_methods.rb:279:in `bind_call'
	from /Users/alex/.gem/ruby/3.3.3/gems/sorbet-runtime-0.5.11465/lib/types/private/methods/_methods.rb:279:in `block in _on_method_added'
	from /Users/alex/src/github.com/Shopify/ruby-lsp-rails/lib/ruby_lsp/ruby_lsp_rails/runner_client.rb:147:in `trigger_reload'
	from /Users/alex/src/github.com/Shopify/ruby-lsp-rails/lib/ruby_lsp/ruby_lsp_rails/addon.rb:117:in `workspace_did_change_watched_files'
	from /Users/alex/.gem/ruby/3.3.3/gems/ruby-lsp-0.17.9/lib/ruby_lsp/server.rb:679:in `block in workspace_did_change_watched_files'
	from /Users/alex/.gem/ruby/3.3.3/gems/ruby-lsp-0.17.9/lib/ruby_lsp/server.rb:679:in `each'
	from /Users/alex/.gem/ruby/3.3.3/gems/ruby-lsp-0.17.9/lib/ruby_lsp/server.rb:679:in `workspace_did_change_watched_files'
	from /Users/alex/.gem/ruby/3.3.3/gems/ruby-lsp-0.17.9/lib/ruby_lsp/server.rb:78:in `process_message'
	from /Users/alex/.gem/ruby/3.3.3/gems/ruby-lsp-0.17.9/lib/ruby_lsp/base_server.rb:128:in `block in new_worker'

```

# Cause

This happens because of this alias:

https://github.com/Shopify/ruby-lsp-rails/blob/d55243dfd3accd8decbe8bf58a6fc96a64ad869d/lib/ruby_lsp/ruby_lsp_rails/runner_client.rb#L184

Which will always point to `RunnerClient#send_message`, regardless of any overrides that might be made by subclasses.

That implementation will crash if `@stdin` is `nil`, such as for `NullClient`.

<details><summary>In-depth explanation</summary>

The `AddOn` starts its life with a `NullClient`, which is intended to do nothing while the application boots:

https://github.com/Shopify/ruby-lsp-rails/blob/d55243dfd3accd8decbe8bf58a6fc96a64ad869d/lib/ruby_lsp/ruby_lsp_rails/addon.rb#L28

When the application is done booting, it'll get replaced with a real `RunnerClient`:

https://github.com/Shopify/ruby-lsp-rails/blob/d55243dfd3accd8decbe8bf58a6fc96a64ad869d/lib/ruby_lsp/ruby_lsp_rails/addon.rb#L32-L36

If you try to call `send_notification` too early, while the `@client` is still a `NullClient`, it'll crash here:

https://github.com/Shopify/ruby-lsp-rails/blob/d55243dfd3accd8decbe8bf58a6fc96a64ad869d/lib/ruby_lsp/ruby_lsp_rails/runner_client.rb#L179

This is surprising, because you would've expected it to call this no-op override instead:

https://github.com/Shopify/ruby-lsp-rails/blob/d55243dfd3accd8decbe8bf58a6fc96a64ad869d/lib/ruby_lsp/ruby_lsp_rails/runner_client.rb#L239-L242

This can be confirmed with a simple log line:

```rb
$stderr.puts("RunnerClient#send_message: #{self.inspect}, #{@stdin.inspect}")
# prints: RunnerClient#send_message: #<RubyLsp::Rails::NullClient:0x000000030b07b058>, nil
```

Sending `send_notification` to a `@client` is resolving to the `RunnerClient` implementation, even though `self` is clearly a `NullClient`.

</details>

# Solution ideas

## Delete `#send_notification`

Just remove the alias, and only call `send_message`

## Use a method instead

Replace the alias with a method definition, which will always dynamically dispatch to the correct override of `send_message`:

```diff
- alias_method :send_notification, :send_message
+ def send_notification(...) = send_message(...)
```

### Redeclare the alias

```diff
class NullClient
  def send_message(request, params = nil) 
    # no-op 
  end
 
+ alias_method :send_notification, :send_message
```

### Use an explicit no-op override

```diff
class NullClient
  def send_message(request, params = nil) 
    # no-op 
  end
 
+ def send_notification(request, params = nil) 
+  # no-op
+ end
```

amomchilov:
Fixes #425

### DIFF
--- a/lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
@@ -170,7 +170,7 @@ module RubyLsp
         read_response
       end
 
-      sig { params(request: String, params: T.nilable(T::Hash[Symbol, T.untyped])).void }
+      sig { overridable.params(request: String, params: T.nilable(T::Hash[Symbol, T.untyped])).void }
       def send_message(request, params = nil)
         message = { method: request }
         message[:params] = params if params
@@ -181,10 +181,11 @@ module RubyLsp
         # The server connection died
       end
 
+      # Notifications are like messages, but one-way, with no response sent back.
       sig { params(request: String, params: T.nilable(T::Hash[Symbol, T.untyped])).void }
       def send_notification(request, params = nil) = send_message(request, params)
 
-      sig { returns(T.nilable(T::Hash[Symbol, T.untyped])) }
+      sig { overridable.returns(T.nilable(T::Hash[Symbol, T.untyped])) }
       def read_response
         headers = @stdout.gets("\r\n\r\n")
         raise IncompleteMessageError unless headers
--- a/lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
@@ -181,7 +181,8 @@ module RubyLsp
         # The server connection died
       end
 
-      alias_method :send_notification, :send_message
+      sig { params(request: String, params: T.nilable(T::Hash[Symbol, T.untyped])).void }
+      def send_notification(request, params = nil) = send_message(request, params)
 
       sig { returns(T.nilable(T::Hash[Symbol, T.untyped])) }
       def read_response
--- a/test/ruby_lsp_rails/runner_client_test.rb
+++ b/test/ruby_lsp_rails/runner_client_test.rb
@@ -87,5 +87,33 @@ module RubyLsp
         FileUtils.mv("test/dummy/config/application.rb.bak", "test/dummy/config/application.rb")
       end
     end
+
+    class NullClientTest < ActiveSupport::TestCase
+      setup { @client = NullClient.new }
+
+      test "#shutdown is a no-op" do
+        assert_nothing_raised { @client.shutdown }
+      end
+
+      test "#stopped? is always true" do
+        assert_predicate @client, :stopped?
+      end
+
+      test "#rails_root is just the current working directory" do
+        assert_equal Dir.pwd, @client.rails_root
+      end
+
+      test "#send_message is a no-op" do
+        assert_nothing_raised { @client.send(:send_message, "request", nil) }
+      end
+
+      test "#send_notification is a no-op" do
+        assert_nothing_raised { @client.send(:send_notification, "request", nil) }
+      end
+
+      test "#read_response is a no-op" do
+        assert_nothing_raised { @client.send(:read_response) }
+      end
+    end
   end
 end


lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
```diff
@@ -170,20 +170,23 @@ def make_request(request, params = nil)
...
         @stdin.write("Content-Length: #{json.length}\r\n\r\n", json)
       rescue Errno::EPIPE
-        # The server connection died
+        $stderr.puts("Lost connection to the Rails server.")
       end
 
-      alias_method :send_notification, :send_message
+      # Notifications are messages that do not expect a response
+      sig { params(request: String, params: T.nilable(T::Hash[Symbol, T.untyped])).void }
+      def send_notification(request, params = nil) = send_message(request, params)
```
amomchilov, line 187: I chose this over the alternatives, so subclasses don't have to remember to add their own alias/override.

It's unlikely that somebody would _not_ want it to just mean the same thing as `send_message`, but if they did, they still have the freedom to override it different, if they would like.

test/ruby_lsp_rails/runner_client_test.rb
```diff
@@ -87,5 +87,33 @@ class RunnerClientTest < ActiveSupport::TestCase
...
+      test "#rails_root is just the current working directory" do
+        assert_equal Dir.pwd, @client.rails_root
+      end
+
+      test "#send_message is a no-op" do
+        assert_nothing_raised { @client.send(:send_message, "request", nil) }
+      end
+
+      test "#send_notification is a no-op" do
+        assert_nothing_raised { @client.send(:send_notification, "request", nil) }
```
amomchilov, line 111: Prior to the fix, this failed with:

```
Error:
RubyLsp::Rails::NullClientTest#test_#send_notification_is_a_no-op:
NoMethodError: undefined method `write' for nil
    lib/ruby_lsp/ruby_lsp_rails/runner_client.rb:180:in `send_message'
    test/ruby_lsp_rails/runner_client_test.rb:111:in `block (2 levels) in <class:NullClientTest>'
    /Users/alex/.gem/ruby/3.3.0/gems/activesupport-7.1.3.3/lib/active_support/testing/assertions.rb:49:in `assert_nothing_raised'
    test/ruby_lsp_rails/runner_client_test.rb:111:in `block in <class:NullClientTest>'
```

lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
```diff
@@ -170,20 +170,23 @@ def make_request(request, params = nil)
         read_response
       end
 
-      sig { params(request: String, params: T.nilable(T::Hash[Symbol, T.untyped])).void }
+      # Messages expect a response back
```
vinistock, line 173: That's not correct. Only requests expect a response back (the method above this one), messages are equivalent to notifications.

lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
```diff
@@ -170,20 +170,23 @@ def make_request(request, params = nil)
...
+      sig { overridable.params(request: String, params: T.nilable(T::Hash[Symbol, T.untyped])).void }
       def send_message(request, params = nil)
         message = { method: request }
         message[:params] = params if params
         json = message.to_json
 
         @stdin.write("Content-Length: #{json.length}\r\n\r\n", json)
       rescue Errno::EPIPE
-        # The server connection died
+        $stderr.puts("Lost connection to the Rails server.")
```
vinistock, line 182: If the connection to the server was lost in the middle of the coding session, this will spam the user's output tab with this message and no context of why the server died.

Do we really need it?

andyw8:
For future safety, would it be better to have `NullClient` and `RunnerClient` implement a `Client` interface, and avoid the inheritance?

lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
vinistock: No, I think `send_message` still makes sense. But the comment doesn't belong here though.

Sending a notification -> send_message (aliased to send_notification to match the spec)
Sending a request -> send_message + read_response

andyw8:
I've removed myself as reviewer for now, you can re-add me on the discussions are resolved.

amomchilov:
@andyw8 I played around with it, but I don't think the result is particularly better. The two are just fairly overlapping. https://github.com/Shopify/ruby-lsp-rails/commit/e3df69ff01dc97d54878ea45453399c1d29414bc

Perhaps there's room to extract a more minimal `Transport` interface, which only has `send_message` and `receive_response`, and then `NullClient` can go away. There would just be `RunnerClient` with a `@transport` that's either `RunnerTransport` or `NullTransport`.

andyw8 (approved):

PR #428, opened 2024-08-06
Bump ruby-lsp dependency to v0.17.11

andyw8:


st0012 (approved):
Since the change was loosening up returning type, it should not cause any runtime error for users? If that's the case, then I'd say it didn't need to bump minor version.

PR #433, opened 2024-08-09
Make client read and write under mutex lock

vinistock:
Since the language server runs requests concurrently under threads, we should only read or write from the pipes under a mutex lock. Otherwise, we risk having mangled input/output, which will break the communication with the server.


st0012 (approved):

vinistock:
It might be related to the `IncompleteMessage` errors we are sometimes seeing. We have the same mutex synchronization in the Ruby LSP when writing/reading from pipes.

PR #435, opened 2024-08-16
Clear the model's association cache before resolving its association

### LINKED ISSUE #420: go-to-definition for new associations doesn't work until Ruby LSP is restarted
If users add a new association, say `belongs_to :user` to the `Post` model, the go-to-definition for `:user` will not work until the LSP server is restarted.

st0012:
Fixes #420 

This PR also fixes the CI failure with Rails main Gemfile.

### DIFF
--- a/lib/ruby_lsp/ruby_lsp_rails/server.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/server.rb
@@ -152,6 +152,8 @@ module RubyLsp
           }
         end
 
+        # Clear the cache so that we get the latest associations.
+        const.clear_reflections_cache
         association_klass = const.reflect_on_association(params[:association_name].intern).klass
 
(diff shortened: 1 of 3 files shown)


lib/ruby_lsp/ruby_lsp_rails/server.rb
```diff
@@ -152,6 +152,8 @@ def resolve_association_target(params)
           }
         end
 
+        # Clear the cache so that we get the latest associations.
+        const.clear_reflections_cache
```
vinistock, line 156: This approach would mean that we essentially never have reflections cached. Do we know how expensive that could be?

Would it be worth trying to only clear the cache based on file modifications, similar to how we handle reloading on schema changes?

st0012:
I don't have time to continue this PR atm so closing it now.

PR #438, opened 2024-08-26
WIP: Server addons

andyw8:
This PR adds support for addons which need to running within a Rails context.

This is being initially aimed a supporting an LSP for Tapioca.

Related:

- https://github.com/Shopify/tapioca/pull/2001
- https://github.com/Shopify/ruby-lsp/pull/2500

Co-authored with @KaanOzkan and @vinistock

### DIFF
--- a/lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
@@ -41,6 +41,7 @@ module RubyLsp
 
       sig { void }
       def initialize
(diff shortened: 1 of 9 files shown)


lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
```diff
@@ -170,6 +174,12 @@ def make_request(request, params = nil)
         read_response
       end
 
+      # Notifications are like messages, but one-way, with no response sent back.
```
andyw8, line 177: Made public so that addons can call it.

PR #442, opened 2024-08-29
Avoid `setpgrp` if `setsid` fails

prashantvithani:
`setsid` throws [`EPERM`](https://man7.org/linux/man-pages/man2/setsid.2.html#ERRORS) if it's attempted for a process that's already a group leader. It is bound to fail if you run`setpgrp` before `setsid`. As a result, the `ruby-lsp` process goes into stopped state and the client waits forever for the booting to finish.


andyw8:
Ping @prashantvithani

prashantvithani:
Hey, sorry for the late response. It seems this change is not accurate. I'll raise another PR with a proper fix/issue report once I'm able to reproduce it with a simple script.

PR #445, opened 2024-08-30
Improve features documentation

st0012:
Closes https://github.com/Shopify/team-ruby-dx/issues/1241

The idea is to keep things simple and prioritize content creation. Therefore, we decided to display features similar to rust-analyzer [does](https://rust-analyzer.github.io/manual.html#features).

- Add `FEATURES.md` to display supported features with demonstration gifs
- Update readme to point to the new features page
- Some minor tweaks 

### Quick Preview

https://github.com/user-attachments/assets/1e9cdd51-15d9-44c1-b4f8-32b899845f85


.gitignore
```diff
@@ -1,5 +1,6 @@
 /.bundle/
 /doc/
+/docs/
```
vinistock, line 3: If you gitignore this directory, our docs will fail to publish. You need to also add the `-f` (force) flag to [this line](https://github.com/Shopify/ruby-lsp-rails/blob/b47335be55558d6a204e94282a6515d4b399a492/.github/workflows/publish_docs.yml#L29).

FEATURES.md
```diff
@@ -0,0 +1,113 @@
...
+Ruby LSP already provides document symbols for Ruby files, such as classes, modules, methods, etc. But the Rails addon
+provides additional document symbols for Rails specific features.
+
+In VS Code, you can open the document symbols view by pressing `Ctrl + Shift + O`.
+
+### Active Record DSLs and Associations
+
+Navigates between Active Record DSLs and Associations using the `Document Symbol` feature.
+
+![Document Symbol for Active Record DSLs and Associations](./ruby-lsp-rails-document-symbol-ar-model.gif)
```
vinistock, line 36: Can we explicitly say callbacks, validations and associations instead of just DSLs? I'm afraid not everyone reading the documentation will have enough context to know what those are.

FEATURES.md
```diff
@@ -0,0 +1,113 @@
...
+
+#### Rails Generate
+
+![Rails Generate](./ruby-lsp-rails-generate-command.gif)
+
+#### Rails Destroy
+
+![Rails Destroy](./ruby-lsp-rails-destroy-command.gif)
+
+### Icon
```
vinistock, line 92: This section being called icon is a bit odd. Would it make sense to move the content directly under `Ruby File Operations`?

FEATURES.md
```diff
@@ -0,0 +1,98 @@
+# Features
+
```
st0012, line 2: I decided not to add table of contents because RDoc generates header anchors in a weird format, like `label-Document+Symbol` for `Document Symbol`, instead of just `document-symbol`.
Since `snapper` already has a table of contents on the side, it should be fine.

vinistock (approved):
Nice

PR #454, opened 2024-09-18
Add support for server addons

andyw8:
Co-authored with @vinistock.

This is initially to support the Tapioca LSP.

There is some tricky logic involving two mutexes that was needed to ensure things are loaded in the correct order.


andyw8:
FYI @KaanOzkan

lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
```diff
@@ -96,6 +96,14 @@ def initialize
         raise InitializationError, @stderr.read
       end
 
+      sig { params(server_addon_path: String).void }
+      def register_server_addon(server_addon_path)
```
KaanOzkan, line 100: Why do we need both `inherited` and `register_server_addon`?

lib/ruby_lsp/ruby_lsp_rails/addon.rb
```diff
@@ -28,20 +28,31 @@ def initialize
...
+          @addon_mutex.lock
+          @client_mutex.synchronize { @rails_runner_client = RunnerClient.create_client }
+          @addon_mutex.unlock
+        end
       end
 
       sig { returns(RunnerClient) }
-      attr_reader :rails_runner_client
+      def rails_runner_client
+        @addon_mutex.synchronize { @rails_runner_client }
```
KaanOzkan, line 44: What happens if requests are sent to NullClient?

lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
andyw8: 🤔 perhaps we don't.... @vinistock ?

lib/ruby_lsp/ruby_lsp_rails/addon.rb
KaanOzkan: I wonder if it's worth holding on to them and forwarding to real client but it's not a priority at all.

lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
andyw8: > We need to do some testing first on a partial rollout.

We can manage that on the Tapioca side though, right?

vinistock: Currently, there's no way to perform a partial rollout. But we could inform addons if experimental features are enabled so that they can decide whether to do something or not.

That would give you some control.

KaanOzkan: As I said above, with this implementation, rollout can be controlled by the individual addons through the trigger of `register_server_addon`. It's not an issue.

PR #464, opened 2024-09-25
Skip directories for view paths

andyw8:
Closes https://github.com/Shopify/ruby-lsp-rails/issues/463

### DIFF
--- a/lib/ruby_lsp/ruby_lsp_rails/code_lens.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/code_lens.rb
@@ -199,8 +199,12 @@ module RubyLsp
           .downcase
 
         view_uris = Dir.glob("#{@client.rails_root}/app/views/#{controller_name}/#{action_name}*").map! do |path|
+          # it's possible we could have a directory with the same name as the action, so we need to skip those
+          next if FileTest.directory?(path)
+
           URI::Generic.from_path(path: path).to_s
-        end
+        end.compact
+
         return if view_uris.empty?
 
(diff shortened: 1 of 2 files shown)


lib/ruby_lsp/ruby_lsp_rails/code_lens.rb
```diff
@@ -199,8 +199,12 @@ def add_jump_to_view(node)
           .downcase
 
         view_uris = Dir.glob("#{@client.rails_root}/app/views/#{controller_name}/#{action_name}*").map! do |path|
+          # it's possible we could have a directory with the same name as the action, so we need to skip those
+          next if FileTest.directory?(path)
+
           URI::Generic.from_path(path: path).to_s
-        end
+        end.compact
```
vinistock, line 206: We should use `filter_map` instead of `map...compact`.

lib/ruby_lsp/ruby_lsp_rails/code_lens.rb
```diff
@@ -199,8 +199,12 @@ def add_jump_to_view(node)
           .downcase
 
         view_uris = Dir.glob("#{@client.rails_root}/app/views/#{controller_name}/#{action_name}*").map! do |path|
+          # it's possible we could have a directory with the same name as the action, so we need to skip those
+          next if FileTest.directory?(path)
```
vinistock, line 203: I've never seen `FileTest` used before. Is this equivalent to `File.directory?` and is there a reason to prefer this one?

PR #469, opened 2024-10-02
Split `ServerAddon` into new file

johansenja:
- ~~It is able to benefit from `strict` typing now~~ removed because of some runtime problems
- It simplifies `server.rb`
- It is now possible to `require` it specifically in other files, eg if you want to write tests for your server addon, you can `require` the server addon file specifically

I originially had some other changes here too, but there are no longer needed!

### DIFF
--- a/lib/ruby_lsp/ruby_lsp_rails/server_addon.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/server_addon.rb
@@ -1,69 +1,53 @@
-# typed: strict
+# typed: false
 # frozen_string_literal: true
-
-require "sorbet-runtime"
+#
+# requiring sorbet-runtime in the server can lead to some problems, so this is untyped
+# https://github.com/Shopify/ruby-lsp-rails/pull/469#issuecomment-2391429546
 
 module RubyLsp
   module Rails
     class ServerAddon
-      extend T::Sig
-      extend T::Helpers
-
-      abstract!
-
-      @server_addon_classes = T.let([], T::Array[T.class_of(ServerAddon)])
-      @server_addons = T.let({}, T::Hash[String, ServerAddon])
-
-      module Types
-        Params = T.type_alias { T::Hash[T.untyped, T.untyped] }
-        Response = T.type_alias { T::Hash[Symbol, T.untyped] }
-      end
+      @server_addon_classes = []
+      @server_addons = {}
 
       class << self
-        extend T::Sig
-
         # We keep track of runtime server add-ons the same way we track other add-ons, by storing
         # classes that inherit from the base one
-        sig { params(child: T::Class[T.anything]).void }
         def inherited(child)
-          @server_addon_classes << T.cast(child, T.class_of(ServerAddon))
+          @server_addon_classes << child
           super
         end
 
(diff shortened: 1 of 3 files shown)


lib/ruby_lsp/ruby_lsp_rails/addon.rb
```diff
@@ -35,7 +35,16 @@ def initialize
         Thread.new do
           @addon_mutex.synchronize do
             # We need to ensure the Rails client is fully loaded before we activate the server addons
-            @client_mutex.synchronize { @rails_runner_client = RunnerClient.create_client }
+            @client_mutex.synchronize do
```
vinistock, line 38: We intentionally don't want to slow down booting the main language server, which is why the client mutex is not locked here.

With the current mechanism, anyone add-on invoking `rails_runner_client` will be stuck until booting the runner client finishes, but it will not block indexing by the main LSP.

We tested this in the Tapioca add-on we're building and this code guarantees correct ordering without blocking the main thread:

```ruby
def activate(global_state, outgoing_queue)
  # get a handle for the addon
  addon = ::RubyLsp::Addon.get("Ruby LSP Rails")

  Thread.new do
    # get a handle to the runner client. This will block this new thread until
    # the Rails add-on finishes initializing and unlocks the client mutex, but
    # the main thread will continue to run properly without stopping indexing
    @rails_runner_client = addon.rails_runner_client
          
    @rails_runner_client.register_server_addon(File.expand_path("server_addon.rb", __dir__))
end
```

You can check the rest of the approach in https://github.com/Shopify/tapioca/pull/2001.

lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
```diff
@@ -7,24 +7,29 @@
 module RubyLsp
   module Rails
     class RunnerClient
+      @instance = T.let(nil, T.nilable(RunnerClient))
+
```
vinistock, line 11: I'm not sure we want to make this a singleton. Other add-ons should only access the client through the `Addon#rails_runner_client` API to ensure proper ordering.

vinistock (changes requested):
I think extracting the server addon definition to another file is fine, but the rails runner client changes are breaking some of the conventions we're putting in place

lib/ruby_lsp/ruby_lsp_rails/addon.rb
johansenja: Ah ok, fair enough! I wasn't aware of `::RubyLsp::Addon.get`, but it should certainly make things a bit easier

lib/ruby_lsp/ruby_lsp_rails/server_addon.rb
```diff
@@ -0,0 +1,54 @@
+# typed: false
```
andyw8, line 1: Since `.rubocop.yml` was updated I think this shouldn't be needed?

vinistock (commented):
This needs a rebase, but the split looks good

johansenja:
Sorry, I've been a bit busy - but no, I don't think I will now that things have moved along

PR #472, opened 2024-10-03
Remove Rails doc DSL hover

vinistock:
Now that the Ruby LSP can display the documentation for most declarations, including DSLs invoked at the class level, the value of having these extra links when hovering is considerably reduced.

Let's remove them and simplify our hover implementation.


st0012 (approved):
Nice 👍

PR #474, opened 2024-10-04
Add API for delegate notifications and requests

vinistock:
Ideally, add-ons shouldn't be calling `send_notification` or `make_request` directly. Let's provide a nicer API.

Also, I changed `params` to be a keyword splat since that makes the API more intuitive.

### DIFF
--- a/lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
@@ -174,6 +174,36 @@ module RubyLsp
         nil
       end
 
+      # Delegates a notification to a server add-on
+      sig { params(server_addon_name: String, request_name: String, params: T.untyped).void }
+      def delegate_notification(server_addon_name:, request_name:, **params)
+        send_notification(
+          "server_addon/delegate",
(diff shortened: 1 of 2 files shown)


lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
```diff
@@ -144,6 +144,36 @@ def route(controller:, action:)
         nil
       end
 
+      # Delegates a notification to a server add-on
+      sig { params(server_addon_name: String, request_name: String, params: T.untyped).void }
+      def delegate_notification(server_addon_name, request_name, **params)
```
KaanOzkan, line 149: I feel like kwargs are nicer since it's documenting the arguments. Wdyt?

KaanOzkan (approved):

lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
vinistock: I don't mind. The LSP can now show signature help, so you see the parameters in the editor anyway.

@andyw8 any opinions?

PR #477, opened 2024-10-04
Add delegate notification API

KaanOzkan:
I went with kwargs for all, it differs from `send_notification` this way but I think it's clearer than positional args.


KaanOzkan:
> I think we can skip adding a test for this.

Thanks, I did test manually.

vinistock:
I added this PR #474, which includes a bit more. Can we move forward with that one?

Issue #468, opened 2024-10-02
Update dummy app to Rails 8


andyw8:
With the full release of Rails 8 coming soon, we should update the app in `test/dummy`.

PR #479, opened 2024-10-09
Update for Rails 8

andyw8:
Closes https://github.com/Shopify/ruby-lsp-rails/issues/468

There are two main things here:

* I re-generated the dummy app with `bundle exec rails plugin new` then did some manual clean-up. There are more files than previously, but I think this is fine since otherwise we would need to clean these up on on every Rails update.

* ~Due to Rails 8 changes, we now have to explictly eager load the routes.~

I verified against code-db which is on Rails 7.2.


.rubocop.yml
```diff
@@ -10,7 +10,7 @@ AllCops:
   NewCops: disable
   SuggestExtensions: false
   Exclude:
-    - "test/dummy/db/**/*.rb"
+    - "test/dummy/**/*"
```
andyw8, line 13: We shouldn't care about styling in the dummy app.

andyw8:
I've marked this as Ready for Review, however there are 2 tests which are failing only in CI and I cannot figure out why 😭 .

test/dummy/Rakefile
```diff
@@ -1,5 +1,3 @@
-# frozen_string_literal: true
-
```
vinistock, line 2: Do we need to remove this?

test/dummy/app/models/application_record.rb
```diff
@@ -1,5 +1,3 @@
-# frozen_string_literal: true
-
```
vinistock, line 2: Same thing, why would we remove this?

test/dummy/config.ru
```diff
@@ -1,6 +1,3 @@
-# typed: strict
-# frozen_string_literal: true
```
vinistock, line 2: Same thing, why would we remove this?

test/dummy/config/database.yml
```diff
@@ -11,15 +11,22 @@ default: &default
...
+# SQLite3 write its data on the local filesystem, as such it requires
+# persistent disks. If you are deploying to a managed service, you should
+# make sure it provides disk persistence, as many don't.
+#
+# Similarly, if you deploy your application as a Docker container, you must
+# ensure the database is located in a persisted volume.
 production:
   <<: *default
-  database: db/production.sqlite3
+  # database: path/to/persistent/storage/production.sqlite3
```
vinistock, line 32: Do we need to switch from `db` to storage? We'll need to purge our existing dbs.

test/dummy/config/environment.rb
```diff
@@ -1,5 +1,3 @@
-# frozen_string_literal: true
-
```
vinistock, line 2: Same thing, why would we remove this?

test/dummy/config/environments/development.rb
```diff
@@ -1,69 +1,69 @@
-# frozen_string_literal: true
-
```
vinistock, line 2: Same thing, why would we remove this?

test/dummy/config/environments/production.rb
```diff
@@ -1,89 +1,89 @@
-# frozen_string_literal: true
-
```
vinistock, line 2: Same thing, why would we remove this?

test/dummy/config/environments/test.rb
```diff
@@ -1,7 +1,3 @@
-# frozen_string_literal: true
-
```
vinistock, line 2: Same thing, why would we remove this?

test/dummy/config/initializers/inflections.rb
```diff
@@ -1,5 +1,3 @@
-# frozen_string_literal: true
-
```
vinistock, line 2: Same thing, why would we remove this?

test/dummy/public/400.html
vinistock, line 1: We should remove the public directory. This is not needed for the dummy test app.

.rubocop.yml
vinistock: Why not?

test/dummy/config/database.yml
andyw8: I figure it's better to since this is the new default, otherwise we'll need to update this manually easy time we re-generate the dummy app.

andyw8:
My original intention was to have keep the dummy app as unchanged as possible, so as to minimize the work needed when regenerating it, but we can stick with the current approach.

vinistock (approved):

andyw8:
Re-verified against code-db and Core with the latest changes.

PR #482, opened 2024-10-16
Fix exception output capturing

andyw8:
Closes https://github.com/Shopify/team-ruby-dx/issues/1297

This ensure the failure output can be displayed in the Ruby LSP output panel.

To 🎩 :

* `dev down` on Core
* Save a model to trigger and RBI update
* Observe that the Ruby LSP output panel shows the appropriate error.

Note: Failures that were previously silent, such as hovering over a ActiveRecord model when the DB isn't set up, will now show in the logs.

### DIFF
--- a/lib/ruby_lsp/ruby_lsp_rails/server.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/server.rb
@@ -15,8 +15,8 @@ module RubyLsp
         @stdout.write("Content-Length: #{json_message.length}\r\n\r\n#{json_message}")
       end
 
-      # Log a debug message to the editor's output
-      def debug_message(message)
+      # Log a message to the editor's output panel
+      def log_message(message)
         $stderr.puts(message)
       end
     end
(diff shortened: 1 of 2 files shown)


lib/ruby_lsp/ruby_lsp_rails/server.rb
```diff
@@ -119,12 +119,17 @@ def execute(request, params)
...
+          ServerAddon.delegate(server_addon_name, request_name, params.except(:request_name, :server_addon_name))
         end
+        request_name = request
+        request_name = "#{params[:server_addon_name]}##{params[:request_name]}" if request == "server_addon/delegate"
+      # Since this is a common problem, we show a specific error message to the user, instead of the full stack trace.
+      rescue ActiveRecord::ConnectionNotEstablished
+        log_message("Request #{request_name} failed because database connection was not established.")
       rescue => e
-        send_message({ error: e.full_message(highlight: false) })
+        log_message("Request #{request_name} failed:\n" + e.full_message(highlight: false))
```
andyw8, line 132: @vinistock I updates this so that we show the request name for both `rescue` clauses.

lib/ruby_lsp/ruby_lsp_rails/server.rb
```diff
@@ -119,12 +119,17 @@ def execute(request, params)
           require params[:server_addon_path]
           ServerAddon.finalize_registrations!(@stdout)
         when "server_addon/delegate"
-          server_addon_name = params.delete(:server_addon_name)
```
andyw8, line 122: We need to avoid deleting the params here since they may be needed in the error message.

PR #494, opened 2024-10-25
Add default and null info on hover

damienlethiec:
We plan to get ride of annotate to only use Ruby LSD in my company. The only missing thing for us is displaying defaults and null information for the model on hover.

This small PR tries to solves this problem.

CleanShot 2024-10-26 at 01 35 52@2x

Please let me know if it needs improvements 🙂

### DIFF
--- a/lib/ruby_lsp/ruby_lsp_rails/hover.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/hover.rb
@@ -67,12 +67,27 @@ module RubyLsp
         ) if schema_file
 
         @response_builder.push(
-          model[:columns].map do |name, type|
+          model[:columns].map do |name, type, default_string, null|
             primary_key_suffix = " (PK)" if model[:primary_keys].include?(name)
-            "**#{name}**: #{type}#{primary_key_suffix}\n"
+            suffixes = []
+            suffixes << "default: #{format_default(default_string, type)}" if default_string
+            suffixes << "not null" unless null
+            suffix_string = " - #{suffixes.join(" - ")}" if suffixes.any?
+            "**#{name}**: #{type}#{primary_key_suffix}#{suffix_string}\n"
           end.join("\n"),
           category: :documentation,
         )
+      end
+
+      sig { params(default_string: String, type: String).returns(String) }
+      def format_default(default_string, type)
+        if type == "boolean"
+          default_string == "true" ? "TRUE" : "FALSE"
+        elsif type == "string"
+          default_string.inspect
+        else
+          default_string
+        end
       end
     end
   end
--- a/test/ruby_lsp_rails/hover_test.rb
+++ b/test/ruby_lsp_rails/hover_test.rb
@@ -10,12 +10,14 @@ module RubyLsp
         expected_response = {
           schema_file: "#{dummy_root}/db/schema.rb",
           columns: [
-            ["id", "integer"],
-            ["first_name", "string"],
-            ["last_name", "string"],
-            ["age", "integer"],
-            ["created_at", "datetime"],
-            ["updated_at", "datetime"],
+            ["id", "integer", nil, false],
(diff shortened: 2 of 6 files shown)


test/ruby_lsp_rails/runner_client_test.rb
```diff
@@ -26,13 +26,14 @@ class RunnerClientTest < ActiveSupport::TestCase
...
-          ["updated_at", "datetime"],
-          ["country_id", "integer"],
+          ["id", "integer", nil, false],
+          ["first_name", "string", "", true],
+          ["last_name", "string", nil, true],
+          ["age", "integer", "0", true],
+          ["created_at", "datetime", nil, false],
+          ["updated_at", "datetime", nil, false],
+          ["country_id", "integer", nil, false],
+          ["active", "boolean", "1", false],
```
damienlethiec, line 36: I honestly do not really get why the default is expected to be "1" here. In my real life app, the defaults returned by the runner is "true" in this case.

lib/ruby_lsp/ruby_lsp_rails/hover.rb
```diff
@@ -67,13 +67,28 @@ def generate_column_content(name)
...
+            suffix_string = " - #{suffixes.join(" - ")}" if suffixes.any?
+            "**#{name}**: #{type}#{primary_key_suffix}#{suffix_string}\n"
           end.join("\n"),
           category: :documentation,
         )
       end
+
+      sig { params(default_string: String, type: String).returns(String) }
+      def format_default(default_string, type)
+        if type == "boolean"
```
andyw8, line 84: (nit) Let's use a `case` here since we're comparing against the same thing in each clause.

lib/ruby_lsp/ruby_lsp_rails/hover.rb
```diff
@@ -67,13 +67,28 @@ def generate_column_content(name)
         ) if schema_file
 
         @response_builder.push(
-          model[:columns].map do |name, type|
+          model[:columns].map do |name, type, default_string, null|
```
andyw8, line 70:
```suggestion
          model[:columns].map do |name, type, default_string, nullable|
```

andyw8, line 70:
```suggestion
          model[:columns].map do |name, type, default_value, null|
```

lib/ruby_lsp/ruby_lsp_rails/hover.rb
```diff
@@ -67,13 +67,28 @@ def generate_column_content(name)
...
+            "**#{name}**: #{type}#{primary_key_suffix}#{suffix_string}\n"
           end.join("\n"),
           category: :documentation,
         )
       end
+
+      sig { params(default_string: String, type: String).returns(String) }
+      def format_default(default_string, type)
+        if type == "boolean"
+          default_string == "true" ? "TRUE" : "FALSE"
```
andyw8, line 85:
```suggestion
          default_string == "true" ? "true" : "false"
```
To match how we write those in Ruby.

andyw8 (changes requested):
Thank you, this looks good, just a few suggestions.

test/ruby_lsp_rails/hover_test.rb
```diff
@@ -88,17 +96,21 @@ class User < ApplicationRecord
 
           [Schema](#{URI::Generic.from_path(path: dummy_root + "/db/schema.rb")})
 
-          **id**: integer (PK)
+          **id**: integer (PK) - not null
```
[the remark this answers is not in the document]
damienlethiec: I agree ! 🙂

damienlethiec:
Thanks for the review @andyw8 . I made the changes 🙂

andyw8 (approved):

PR #495, opened 2024-10-26
Update index enhancement due to ruby-lsp changes

andyw8:
Due to https://github.com/Shopify/ruby-lsp/pull/2754


lib/ruby_lsp/ruby_lsp_rails/addon.rb
```diff
@@ -53,7 +53,7 @@ def activate(global_state, outgoing_queue)
         @outgoing_queue << Notification.window_log_message("Activating Ruby LSP Rails add-on v#{VERSION}")
 
         register_additional_file_watchers(global_state: global_state, outgoing_queue: outgoing_queue)
-        @global_state.index.register_enhancement(IndexingEnhancement.new)
+        @global_state.index.register_enhancement(IndexingEnhancement.new(@global_state.index))
```
andyw8, line 56: TBC: https://github.com/Shopify/ruby-lsp/pull/2754/files#r1817851910

lib/ruby_lsp/ruby_lsp_rails/indexing_enhancement.rb
```diff
@@ -3,13 +3,17 @@
...
-    class IndexingEnhancement
+    class IndexingEnhancement < RubyIndexer::Enhancement
       extend T::Sig
-      include RubyIndexer::Enhancement
+
+      sig { params(index: RubyIndexer::Index).void }
+      def initialize(index)
+        super
+        @index = index
+      end
```
vinistock, line 13: We don't need to override initialize. This is already the same in the [parent class](https://github.com/Shopify/ruby-lsp/blob/4c7659be56ab83dae65c3bf63e666b95b684f046/lib/ruby_indexer/lib/ruby_indexer/enhancement.rb#L12).

vinistock (approved):
Just needs to remove the initialize, but the rest looks good

PR #498, opened 2024-10-30
Return `nil` on empty response

vinistock:
We are currently raising if something is printed to stdout from the server and it results in an empty message, but I'm not sure why we were raising in the first place.

This keeps filling up our telemetry with errors that aren't actionable. It could be that the application printed something, but since there's no content, I think it's okay to just return `nil` (since this method is already expected to be able to return `nil`).

### DIFF
--- a/test/ruby_lsp_rails/runner_client_test.rb
+++ b/test/ruby_lsp_rails/runner_client_test.rb
@@ -87,17 +87,16 @@ module RubyLsp
         junk = %{\nputs "1\r\n\r\nhello"}
         File.write("test/dummy/config/application.rb", content + junk)
 
-        capture_subprocess_io do
-          outgoing_queue = Thread::Queue.new
-          client = RunnerClient.create_client(outgoing_queue)
+        outgoing_queue = Thread::Queue.new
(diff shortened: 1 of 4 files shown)


test/ruby_lsp_rails/runner_client_test.rb
```diff
@@ -87,17 +87,20 @@ class RunnerClientTest < ActiveSupport::TestCase
...
+
         capture_subprocess_io do
-          outgoing_queue = Thread::Queue.new
           client = RunnerClient.create_client(outgoing_queue)
-
-          response = T.must(client.model("User"))
-          assert(response.key?(:columns))
-        ensure
-          T.must(outgoing_queue).close
+          response = client.model("User")
```
vinistock, line 94: This test was passing, but the behaviour is actually always incorrect. `client.model` was raising `EmptyMessageError` before, which means we never got to the point of asserting anything and the ensure block saved the test from erroring.

I'm investigating why this test breaks.

vinistock:
For posterity, the issue with Windows was that we were only turning on `sync` for the STD pipes on the server side, but not on the client side, which may result in the OS deciding to buffer and then the client never receives responses back.

PR #501, opened 2024-10-31
Reload routes when changed

andyw8:


test/ruby_lsp_rails/server_test.rb
```diff
@@ -176,10 +176,34 @@ def resolve_route_info(requirements)
...
+  end
+
   private
 
   def response
-    _headers, content = @stdout.string.split("\r\n\r\n")
-    JSON.parse(content, symbolize_names: true)
+    headers, content = @stdout.string.split("\r\n\r\n")
+    content_length = headers[/Content-Length: (\d+)/i, 1].to_i
+    # binding.break
```
andyw8, line 205: @vinistock I'm trying to figure out what's happening here: If I add the breakpoint, and call `@stdout.string = ""` each time, it passes. But I call it from the code it fails with `write': not opened for writing (IOError)`.

lib/ruby_lsp/ruby_lsp_rails/addon.rb
```diff
@@ -119,10 +119,13 @@ def create_definition_listener(response_builder, uri, node_context, dispatcher)
 
       sig { params(changes: T::Array[{ uri: String, type: Integer }]).void }
       def workspace_did_change_watched_files(changes)
-        if changes.any? do |change|
-             change[:uri].end_with?("db/schema.rb") || change[:uri].end_with?("structure.sql")
-           end
+        schema = lambda { |change| change[:uri].end_with?("db/schema.rb") || change[:uri].end_with?("structure.sql") }
+        routes = lambda { |change| change[:uri].end_with?("routes.rb") || change[:uri].match?("routes/**/*.rb") }
```
andyw8, line 123: Refactored to simplify the conditional.

lib/ruby_lsp/ruby_lsp_rails/addon.rb
```diff
@@ -141,7 +144,7 @@ def register_additional_file_watchers(global_state:, outgoing_queue:)
                 register_options: Interface::DidChangeWatchedFilesRegistrationOptions.new(
                   watchers: [
                     Interface::FileSystemWatcher.new(
-                      glob_pattern: "**/*structure.sql",
+                      glob_pattern: "**/*structure.sql,**/*routes.rb,routes/**/*.rb",
```
andyw8, line 147: I think this should almost almost any way that people set up the Rails routes, but if this is a risk of a false positive if someone is using the term `routes` for something else in their app.

andyw8: Possibly:
```
dummy(dev)> Rails.application.routes.draw_paths.map(&:to_s)
=> 
["/Users/andyw8/.gem/ruby/3.3.0/gems/actionview-8.0.0.beta1/config/routes",
 "/Users/andyw8/.gem/ruby/3.3.0/gems/activestorage-8.0.0.beta1/config/routes",
 "/Users/andyw8/.gem/ruby/3.3.0/gems/actioncable-8.0.0.beta1/config/routes",
 "/Users/andyw8/.gem/ruby/3.3.0/gems/actionmailbox-8.0.0.beta1/config/routes",
 "/Users/andyw8/.gem/ruby/3.3.0/gems/actiontext-8.0.0.beta1/config/routes",
 "/Users/andyw8/src/github.com/Shopify/ruby-lsp-rails/test/dummy/config/routes",
 "/Users/andyw8/src/github.com/Shopify/ruby-lsp-rails/test/dummy/config/routes"]
```

PR #503, opened 2024-11-01
Add support for checking for pending migrations and running them

vinistock:
Add the ability to check for pending migrations and run them through the Rails runner client.

For checking if something is pending, we use the Rails API directly, which raises if there are any pending migrations.

For running them, we can't do it in process. The code that runs migrations is not designed to be invoked repeatedly and uses `load`, which means that we could end up with a memory bloat. I used `Open3.capture2` to run in a child process and grab the results.

### DIFF
--- a/lib/ruby_lsp/ruby_lsp_rails/server.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/server.rb
@@ -2,6 +2,7 @@
 # frozen_string_literal: true
 
 require "json"
+require "open3"
 
 # NOTE: We should avoid printing to stderr since it causes problems. We never read the standard error pipe from the
 # client, so it will become full and eventually hang or crash. Instead, return a response with an `error` key.
@@ -109,6 +110,10 @@ module RubyLsp
           send_message(resolve_database_info_from_model(params.fetch(:name)))
         when "association_target_location"
           send_message(resolve_association_target(params))
+        when "pending_migrations_message"
+          send_message({ result: { pending_migrations_message: pending_migrations_message } })
+        when "run_migrations"
+          send_message({ result: run_migrations })
         when "reload"
           ::Rails.application.reloader.reload!
         when "route_location"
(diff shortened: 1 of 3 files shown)


test/ruby_lsp_rails/server_test.rb
```diff
@@ -176,6 +176,28 @@ def resolve_route_info(requirements)
     assert_equal("Hello\n", stderr)
   end
 
+  test "checking for pending migrations" do
+    capture_subprocess_io do
+      system("bundle exec rails g migration CreateStudents name:string")
+    end
+
+    @server.execute("pending_migrations", {})
+    message = response.dig(:result, :pending_migrations)
```
st0012, line 185: Since `pending_migrations` doesn't contain a list of pending migration names but just a message, I wonder if we can skip this namespace? I find it not too useful but a bit misleading instead.

lib/ruby_lsp/ruby_lsp_rails/server.rb
```diff
@@ -252,6 +257,26 @@ def active_record_model?(const)
...
+      rescue ActiveRecord::PendingMigrationError => e
+        e.message
+      end
+
+      def run_migrations
+        # Running migrations invokes `load` which will repeatedly load the same files. It's not designed to be invoked
+        # multiple times within the same process. To avoid any memory bloat, we run migrations in a separate process
+        stdout, status = Open3.capture2(
+          { "VERBOSE" => "true" },
+          "bundle exec rake db:migrate",
```
andyw8, line 275: I expect we should be using `rails db:migrate` to match the modern way.

(it's been available since https://github.com/rails/rails/pull/22288 so there shouldn't be any problem with backwards compatibility).

andyw8 (approved):

lib/ruby_lsp/ruby_lsp_rails/server.rb
vinistock: Changed to `rails db:migrate` 👍

PR #509, opened 2024-11-05
Remove usage of present? from hover

vinistock:
We cannot use `present?` in the add-on side of the code because that method is defined by `activesupport` and that gem is not required on the LSP side.

It's only safe to use it in the runtime server side. This is currently breaking hover.

Note: tests won't catch this because in tests the Rails context is loaded through the dummy application, which makes tests pass despite this code breaking every time in real scenarios.

PR #510, opened 2024-11-05
Avoid loading Rails or ActiveSupport in tests

andyw8:
To prevent issues such as https://github.com/Shopify/ruby-lsp-rails/pull/509 we can switch to using `Minitest::Test` instead of `ActiveSupport::TestCase` for the test files.

I have temporarily added the `test_declarative` gem to avoid editing all the test declarations, but if we decide to go ahead then I can do that.

We have one test file (`server_test.rb`) that **does** need to load Rails so I have split that out.

~There is currently one failing test that I need to look into.~ fixed

### DIFF
--- a/test/ruby_lsp_rails/runner_client_test.rb
+++ b/test/ruby_lsp_rails/runner_client_test.rb
@@ -9,11 +9,11 @@ module RubyLsp
     class RunnerClientTest < Minitest::Test
       def setup
         @outgoing_queue = Thread::Queue.new
-        @client = T.let(RunnerClient.new(@outgoing_queue), RunnerClient)
+        @client = T.let(RunnerClient.new(@outgoing_queue), T.nilable(RunnerClient))
       end
 
       def teardown
-        @client.shutdown
+        T.must(@client).shutdown
 
         # On Windows, the server process sometimes takes a lot longer to shutdown and may end up getting force killed,
         # which makes this assertion flaky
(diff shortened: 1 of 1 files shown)


andyw8:
@vinistock any thoughts?

test/test_helper_server.rb
```diff
@@ -0,0 +1,11 @@
+# typed: true
+# frozen_string_literal: true
+
+# Configure Rails Environment
+ENV["RAILS_ENV"] = "test"
+
+require "test_helper"
+
+require_relative "../test/dummy/config/environment"
```
vinistock, line 9: I think this is the key culprit here. I don't even think we need to switch fully to just Minitest.

As long as we require only the parts of Active Support related to testing, we can continue with the same setup.

But we can't load the dummy app for every test because that will load everything and get us into trouble.

Rakefile
```diff
@@ -13,7 +13,13 @@ require "rake/testtask"
...
 end
 
-task default: [:"db:setup", :test]
+Rake::TestTask.new("test:server") do |t|
+  t.libs << "test"
+  t.libs << "lib"
+  t.test_files = ["test/ruby_lsp_rails/server_test.rb"]
+end
+
+task default: [:"db:setup", :test, "test:server"]
```
vinistock, line 25: Did you test if adding a `present?` in the add-on code breaks when running `bundle exec rake`?

I _think_ because `test:server` is last, then the related test helper only gets required at the end, but I'm not 100% sure.

```suggestion
task default: ["db:setup", :test, "test:server"]
```

PR #508, opened 2024-11-05
Upgrade gems and remove webmock

vinistock:
We are currently a bit behind on gem upgrades. This PR upgrades a bunch of them and also removes `webmock`, which we don't use. If we ever need it, we can bring it back.

PR #511, opened 2024-11-05
Remove remaining traces of Webmock

andyw8:
Following up from https://github.com/Shopify/ruby-lsp-rails/pull/508


sorbet/rbi/gems/minitest@5.25.1.rbi
```diff
@@ -1346,18 +1346,6 @@ class Minitest::Test < ::Minitest::Runnable
   include ::Mocha::API
   extend ::Minitest::Guard
 
-  # source://webmock/3.23.1lib/webmock/minitest.rb#29
```
andyw8, line 1349: (regenerated)

Issue #512, opened 2024-11-06
Plugin crashes on Rails < 7.1 because `ActiveRecord::Migration#check_all_pending!` isn't present in those versions


moskyb:
Since the merge of https://github.com/Shopify/ruby-lsp-rails/pull/503, this plugin crashes the entire LSP in my VSCode setup, with the following error:

```
2024-11-06 11:40:50.977 [info] (myrepo) Request  failed:
2024-11-06 11:40:50.977 [info] (myrepo) /Users/ben/.rbenv/versions/3.3.5/lib/ruby/gems/3.3.0/gems/activerecord-7.0.8.6/lib/active_record/migration.rb:964:in `block in method_missing': undefined method `check_all_pending!' for an instance of ActiveRecord::Migration (NoMethodError)
```

This is because my rails app is running 7.0.8.6, which does not have the method `ActiveRecord::Migration#check_all_pending!` - it was added in [Rails 7.1](https://github.com/rails/rails/blob/7-1-stable/activerecord/CHANGELOG.md).

Related: https://github.com/Shopify/ruby-lsp-rails/issues/500, as it would likely have shown this issue prior to merge.

wamazing-sei-sato:
Hi team,

First, I’d like to thank the maintainers of `ruby-lsp-rails` for their work on this valuable tool. I wanted to mention that I'm also experiencing this compatibility issue in a project using Rails v7.0.3, with the same error and LSP crash as described in #512.

As a temporary workaround, I updated my `Gemfile` to lock `ruby-lsp-rails` to an earlier version, `v0.3.21`, which resolved the issue for my setup:

```ruby
group :development do
  gem 'ruby-lsp-rails', '0.3.21'
  ...
end
```

I hope this information is helpful. Thanks again to the team for their efforts on this project.

andyw8:
Shipped: https://rubygems.org/gems/ruby-lsp-rails/versions/0.3.24

PR #513, opened 2024-11-06
Don't call `check_all_pending!` for older Rails

andyw8:
Closes https://github.com/Shopify/ruby-lsp-rails/issues/512


flemaitre33:
Hello @vinistock could you publish a new version for VSCode with this fix please 🙏  ?

Issue #516, opened 2024-11-07
Starting ruby-lsp-rails fails if database doesn't exist


andyw8:
In https://github.com/Shopify/ruby-lsp-rails/pull/503 we added a check for pending migrations when starting ruby-lsp-rails.

If the database doesn't exist, this results in a failure which causes Ruby LSP to stop responding.

On code-db (an internal Shopify app) I can reproduce by running `rake db:drop` then starting Ruby LSP.

```2024-11-07 10:00:53.837 [info] (code-db) Finished booting Ruby LSP Rails server
2024-11-07 10:00:53.844 [info] (code-db) Request  failed:

2024-11-07 10:00:53.844 [info] (code-db) /Users/andyw8/.gem/ruby/3.3.1/gems/activerecord-7.2.2/lib/active_record/connection_adapters/trilogy_adapter.rb:34:in `rescue in new_client': We could not find your database: code_db_development. Available database configurations can be found in config/database.yml. (ActiveRecord::NoDatabaseError)
```

PR #517, opened 2024-11-07
Improve error handling for database problems

andyw8:
Closes https://github.com/Shopify/ruby-lsp-rails/issues/516

This change allows Ruby LSP to stay alive even if an error occurred when trying to connect to the database.

I also added a specific `rescue` for `ActiveRecord::NoDatabaseError` since this could be a common problem. Without that, users would see a very long stacktrace in the logs.

It may _partially_ help with https://github.com/Shopify/ruby-lsp/issues/2838 too.

### DIFF
--- a/lib/ruby_lsp/ruby_lsp_rails/server.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/server.rb
@@ -19,6 +19,7 @@ module RubyLsp
       # Log a message to the editor's output panel
       def log_message(message)
         $stderr.puts(message)
+        send_message({ result: nil })
       end
     end
 
@@ -103,6 +104,9 @@ module RubyLsp
       end
 
       def execute(request, params)
+        request_name = request
+        request_name = "#{params[:server_addon_name]}##{params[:request_name]}" if request == "server_addon/delegate"
+
         case request
         when "shutdown"
           @running = false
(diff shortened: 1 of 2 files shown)


lib/ruby_lsp/ruby_lsp_rails/server.rb
```diff
@@ -133,8 +134,10 @@ def execute(request, params)
       # Since this is a common problem, we show a specific error message to the user, instead of the full stack trace.
       rescue ActiveRecord::ConnectionNotEstablished
         log_message("Request #{request_name} failed because database connection was not established.")
+      rescue ActiveRecord::NoDatabaseError
+        log_message("Request #{request_name} failed because the database does not exist.")
       rescue => e
-        log_message("Request #{request_name} failed:\n" + e.full_message(highlight: false))
+        log_message("Request #{request} failed:\n #{e.class}" + e.full_message(highlight: false))
```
andyw8, line 140: `request_name` was wrong here.

lib/ruby_lsp/ruby_lsp_rails/server.rb
```diff
@@ -19,6 +19,7 @@ def send_message(message)
       # Log a message to the editor's output panel
       def log_message(message)
         $stderr.puts(message)
+        send_message({ result: nil })
```
vinistock, line 22: This is indeed the right solution as we need to return something even when errors occur. But this is not the right place to put it.

The goal of the `log_message` method is to allow both the server and its add-ons to send debug messages to the output tab, even if disconnected from providing a result.

If a database error was to occur in a notification, where no response is expected, then we'd be writing a nil result to the pipe, which wouldn't be read by the notification. The next request would then incorrectly read it and we'd end up with the request/response cycle out of sync.

We need to move this line to where error handling happens, so that `log_message` is not coupled with returning responses to the editor. And this result can only be returned when rescuing errors for requests - not for notifications.

andyw8: Ah yes, I wasn't considering usage from add-ons.

https://github.com/Shopify/ruby-lsp-rails/pull/517

Issue #519, opened 2024-11-07
Improve diagnostics for situations where Rails fails to boot


andyw8:
Found while looking into https://github.com/Shopify/ruby-lsp-rails/issues/514

If the app's `config/database.yml` is missing, the add-onswill fail to start in a way where it's hard to see the underlying problem:

```
2024-11-07 13:24:06.384 [info] (ruby-lsp-rails) [Error - 1:24:06 PM] Ruby LSP Rails failed to initialize server: /Users/andyw8/src/github.com/Shopify/ruby-lsp-rails/lib/ruby_lsp/ruby_lsp_rails/runner_client.rb:296:in `block in read_response': RubyLsp::Rails::RunnerClient::EmptyMessageError (RubyLsp::Rails::RunnerClient::EmptyMessageError)
	from /Users/andyw8/src/github.com/Shopify/ruby-lsp-rails/lib/ruby_lsp/ruby_lsp_rails/runner_client.rb:293:in `synchronize'
	from /Users/andyw8/src/github.com/Shopify/ruby-lsp-rails/lib/ruby_lsp/ruby_lsp_rails/runner_client.rb:293:in `read_response'
...
...
```

Although this particular situation is probably uncommon, I expect that other kinds of problems may result in similar output.

PR #520, opened 2024-11-07
Improve diagnostics for situations where Rails fails to boot

andyw8:
Closes https://github.com/Shopify/ruby-lsp-rails/issues/519

### DIFF
--- a/lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
@@ -109,7 +109,7 @@ module RubyLsp
           end,
           Thread,
         )
-      rescue Errno::EPIPE, IncompleteMessageError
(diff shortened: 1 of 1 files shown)


lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
```diff
@@ -109,7 +109,7 @@ def initialize(outgoing_queue)
           end,
           Thread,
         )
-      rescue Errno::EPIPE, IncompleteMessageError
+      rescue Errno::EPIPE, IncompleteMessageError, MessageError
```
andyw8, line 112: (Perhaps we should rescue _any_ `StandardError` here?)

KaanOzkan: That sounds good to me.

PR #523, opened 2024-11-08
Fix log message flow

andyw8:
Addresses https://github.com/Shopify/ruby-lsp-rails/pull/517/files#r1834925776

### DIFF
--- a/lib/ruby_lsp/ruby_lsp_rails/server.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/server.rb
@@ -19,7 +19,6 @@ module RubyLsp
       # Log a message to the editor's output panel
       def log_message(message)
         $stderr.puts(message)
-        send_message({ result: nil })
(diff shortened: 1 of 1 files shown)


lib/ruby_lsp/ruby_lsp_rails/server.rb
```diff
@@ -135,10 +134,13 @@ def execute(request, params)
       # Since this is a common problem, we show a specific error message to the user, instead of the full stack trace.
       rescue ActiveRecord::ConnectionNotEstablished
         log_message("Request #{request_name} failed because database connection was not established.")
+        send_message({ result: nil })
```
andyw8, line 137: 🤔 Perhaps this should be in `ensure`?

andyw8:
Resolved by https://github.com/Shopify/ruby-lsp-rails/pull/527

PR #527, opened 2024-11-18
Unify all error handling for the server

vinistock:
Closes https://github.com/Shopify/ruby-lsp/issues/2886

We cannot use blanket rescue statements for the entire server because the behaviour for handling errors is different between requests and notifications.

For requests, we want to return `error: message` instead of the typicaly `result: something`. And for notifications, we cannot return anything, we can only log that an error occurred.

The approach that I took here is to create handlers that we can use both ourselves and in server add-ons, that provide appropriate error handling for requests or notifications.

I also created convenience  methods for returning responses or errors.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,7 @@
+{
+    "rubyLsp.indexing": {
+        "includedPatterns": [
+            "test/dummy/**/*.rb"
+        ]
+    }
+}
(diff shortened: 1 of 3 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,7 @@
+{
+    "rubyLsp.indexing": {
+        "includedPatterns": [
+            "test/dummy/**/*.rb"
+        ]
+    }
```
vinistock, line 6: We need to ensure that the dummy files are indexed to be able to test our own features. A recent optimization in the LSP made us fully ignore the entire test directory straight away.

KaanOzkan (approved):
I really like the { result: nil } clean up

Issue #525, opened 2024-11-14
Handle `class_methods do` blocks for concerns


andyw8:
(details to be added)

cc @toddkummer

andyw8:
@toddkummer this shipped in v0.3.27, if you want to check it's working wth your gem.

PR #530, opened 2024-11-19
Add support for `class_methods do` blocks within concerns

andyw8:
Closes https://github.com/Shopify/ruby-lsp-rails/issues/525

Depends on https://github.com/Shopify/ruby-lsp-rails/pull/529

Here are the new places in Rails we now find:

```
ActionMailbox::Callbacks
ActionMailbox::Routing
ActionMailer::Rescuable
ActionText::Attachable
ActionText::Attachments::Minification
ActionText::Attachments::TrixConversion
ActionText::Attribute
ActionText::Rendering
ActionText::Serialization
ActionText::Serialization
ActiveRecord::ConnectionAdapters::ColumnMethods
ActiveRecord::Encryption::Configurable
ActiveRecord::Encryption::Contexts
ActiveRecord::Encryption::EncryptableRecord
ActiveRecord::Encryption::ExtendedDeterministicQueries::CoreQueries
ActiveStorage::Attached::Model
Rails::Command::Behavior
```

### DIFF
--- a/test/ruby_lsp_rails/indexing_enhancement_test.rb
+++ b/test/ruby_lsp_rails/indexing_enhancement_test.rb
@@ -8,7 +8,7 @@ module RubyLsp
     class IndexingEnhancementTest < ActiveSupport::TestCase
       class << self
         # For these tests, it's convenient to have the index fully populated with Rails information, but we don't have
-        # to reindex on every single example or that will be too slow
+        # to re-index on every single example or that will be too slow
         def populated_index
           @index ||= begin
             index = RubyIndexer::Index.new
@@ -20,17 +20,17 @@ module RubyLsp
 
       def setup
         @index = self.class.populated_index
+        @indexable_path = RubyIndexer::IndexablePath.new(nil, "/fake.rb")
       end
 
       def teardown
         # Prevent state leaking between tests
-        indexable_path = RubyIndexer::IndexablePath.new(nil, "/fake.rb")
-        @index.delete(indexable_path)
+        @index.delete(@indexable_path)
         @index.instance_variable_set(:@ancestors, {})
       end
 
       test "ClassMethods module inside concerns are automatically extended" do
-        @index.index_single(RubyIndexer::IndexablePath.new(nil, "/fake.rb"), <<~RUBY)
+        @index.index_single(@indexable_path, <<~RUBY)
           module Verifiable
             extend ActiveSupport::Concern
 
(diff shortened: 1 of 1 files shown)


vinistock:
The code looks good to me. Not sure about the test failures, but after a rebase it might be easier to understand what's going on.

test/ruby_lsp_rails/indexing_enhancement_test.rb
```diff
@@ -7,14 +7,12 @@ module RubyLsp
   module Rails
     class IndexingEnhancementTest < ActiveSupport::TestCase
       class << self
-        # For these tests, it's convenient to have the index fully populated with Rails information, but we don't have
-        # to reindex on every single example or that will be too slow
+        # For these tests, it's convenient to have the index fully populated with Rails information.
+        # Reindexing each time is slow, but prevents state leaking between tests.
         def populated_index
-          @index ||= begin
```
andyw8, line 13: @vinistock avoiding the memoization solves in test instability here. It's obviously slower though.

A few possible alternatives:
* Have each test be responsible to removing things it added to the index (maybe compare the state before and after in a `teardown` to ensure clean-up isn't forgotten)
* Have some naming convention for things we use in tests, and have them automatically removed from the index in a `teardown`.
* Have a way to skip indexing of RBS in `index_all` (which takes a large portion of the test time).

vinistock: How about we just add note to the test and use different names in each example? Trying to clean up all of the index state might be difficult and it will require exposing some internal APIs.

lib/ruby_lsp/ruby_lsp_rails/indexing_enhancement.rb
```diff
@@ -6,6 +6,13 @@ module Rails
     class IndexingEnhancement < RubyIndexer::Enhancement
       extend T::Sig
 
+      sig { params(listener: RubyIndexer::DeclarationListener).void }
+      def initialize(listener)
+        super
+        # We need this to prevent Sorbet from complaining that @listener is undeclared
+        @listener = listener
+      end
```
vinistock, line 14: ✂️

test/ruby_lsp_rails/indexing_enhancement_test.rb
```diff
@@ -22,16 +22,71 @@ def setup
         @index = self.class.populated_index
       end
 
+      def teardown
+        # Prevent state leaking between tests
+        indexable_path = RubyIndexer::IndexablePath.new(nil, "/fake.rb")
```
vinistock, line 27: Let's make this an instance variable in `setup` and then use it both for deleting and for indexing since this is the same on every example.

PR #548, opened 2024-12-17
Extract TestUtils

andyw8:
I'm considering extracting this so that we can use it from https://github.com/Shopify/tapioca/pull/2108

cc @KaanOzkan


st0012:
They don't seem to be tied to this addon or Rails. Maybe it can just be part of `ruby-lsp`'s test helper?

PR #549, opened 2024-12-17
Bump ruby-lsp dependency to v0.22.1

andyw8:
We will have v0.23 of Ruby LSP out soon but let's do this first as a smaller step.


andyw8:
I just wanted to verify we hadn't introduced any accidental breakages. We can skip it.

PR #550, opened 2024-12-17
Prepare for Ruby LSP v0.23

andyw8:


Gemfile
```diff
@@ -19,7 +19,8 @@ gem "tapioca", "~> 0.13", require: false, platforms: :ruby
 gem "psych", "~> 5.1", require: false
 gem "rails", "8.0.0"
 
-platforms :mingw, :x64_mingw, :mswin, :jruby do
```
andyw8, line 22: Investigating some odd failures when this is present:

```
Error:
RubyLsp::Rails::CodeLensTest#test_ignores_unnamed_tests_(empty_string):
Gem::MissingSpecError: Could not find 'tzinfo-data' (>= 0) among 145 total gem(s)
Checked in 'GEM_PATH=/Users/andyw8/.gem/ruby/3.3.0:/opt/rubies/3.3.0/lib/ruby/gems/3.3.0' , execute `gem env` for more information
```

andyw8: Seems to relate to https://github.com/Shopify/ruby-lsp/pull/2937

lib/ruby_lsp/ruby_lsp_rails/definition.rb
```diff
@@ -145,7 +145,7 @@ def collect_definitions(name)
 
         methods.each do |target_method|
           location = target_method.location
-          file_path = target_method.file_path
+          file_path = T.must(target_method.file_path)
```
vinistock, line 148: Let's improve this entire `each` block. Since we now store the URIs in entries, the code can be simplified. There's no need to instantiate a new URI. Additionally, using the URI directly means this code will work when we start indexing unsaved files with no modifications.

Also, we forgot to use our own helpers to create the range.

```ruby
methods.each do |target_method|
  @response_builder << Interface::Location.new(
    uri: target_method.uri,
    range: range_from_location(target_method.location)
  )
end
```

vinistock (approved):

lib/ruby_lsp/ruby_lsp_rails/definition.rb
andyw8: Updated

PR #551, opened 2024-12-20
WIP: Use test helpers from Ruby LSP

andyw8:
Waiting on https://github.com/Shopify/ruby-lsp/pull/2991 to be released.


andyw8:
Added to https://github.com/Shopify/ruby-lsp-rails/pull/550

PR #552, opened 2025-01-07
Use Ruby 3.4

andyw8:


Gemfile.lock
```diff
@@ -268,10 +278,16 @@ GEM
     zeitwerk (2.7.1)
 
 PLATFORMS
+  aarch64-linux
```
andyw8, line 281: These extra platforms are because Bundler suggested running `bundle lock --normalize-platforms`.

PR #555, opened 2025-01-14
Reduce file watching overhead from Rails

vinistock:
I noticed that we were spending a lot of time doing file IO operations coming from the listen gem. After some investigation, these appear to be coming from Rails' own file watching mechanisms, which spawn threads that will check if views are being modified to trigger rendering or reloading in the background.

We currently don't have any functionality that depends on views being rendered. Additionally, the LSP already watches files, so even if we wanted to trigger view rendering, the language server would be in a better position to dictate which views to reload.

This PR shuts down view file watching, which eliminates a good chunk of overhead that we were seeing in the Tapioca add-on.

### Approach

I initially tried setting `config.file_watcher` to a no-op implementation, but that doesn't work. The reason is because, by the time we start running `server.rb`, Rails has already registered the file watchers and its callbacks, so changing the config won't have any real effect.

The approach I took is to clear the file system hooks array from the Action View path registry. Those hooks are the ones that build the file watchers, spawning listen threads. By clearing that array, I can see the overhead disappear in my profiles.

### DIFF
--- a/lib/ruby_lsp/ruby_lsp_rails/server.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/server.rb
@@ -4,9 +4,6 @@
 require "json"
 require "open3"
 
-# NOTE: We should avoid printing to stderr since it causes problems. We never read the standard error pipe from the
-# client, so it will become full and eventually hang or crash. Instead, return a response with an `error` key.
-
 module RubyLsp
   module Rails
     module Common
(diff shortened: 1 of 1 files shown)


vinistock:
* **#555** <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp-rails/555?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp-rails/555?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

KaanOzkan (approved):

PR #563, opened 2025-01-28
Show progress when reloading the Rails application

vinistock:
One of the slowest parts of DSL RBI generation in the Tapioca add-on is actually the fact that we're reloading the Rails app.

In addition to adding progress notifications there, let's also show the progress of reloading the app, so that the user knows what's going on.

### DIFF
--- a/lib/ruby_lsp/ruby_lsp_rails/server.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/server.rb
@@ -258,8 +258,10 @@ module RubyLsp
             send_result(run_migrations)
           end
         when "reload"
-          with_notification_error_handling(request) do
-            ::Rails.application.reloader.reload!
+          with_progress("rails-reload", "Reloading Ruby LSP Rails instance") do
+            with_notification_error_handling(request) do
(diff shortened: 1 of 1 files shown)


lib/ruby_lsp/ruby_lsp_rails/server.rb
```diff
@@ -258,8 +258,10 @@ def execute(request, params)
             send_result(run_migrations)
           end
         when "reload"
-          with_notification_error_handling(request) do
-            ::Rails.application.reloader.reload!
+          with_progress("rails-reload", "Reloading Rails application context") do
```
andyw8, line 261:
```suggestion
          with_progress("rails-reload", "Reloading Rails application") do
```
just as clear?

vinistock: I was afraid that saying "Reloading application" could make the user think we're reloading the Rails server they might be running, when in fact we're reloading our own custom one.

andyw8: Maybe "Reloading the Ruby LSP Rails instance of your application"

PR #564, opened 2025-01-28
Wrap stderr in a structure log delegator

vinistock:
After #559, the server and its add-ons can send notifications, but regular `puts` that would normally be redirected to stderr are now lost.

Trying to account for any type of print in the notifier thread would be difficult. We need to be able to match against the `\r\n\r\n` separator, so that we can know exactly the start and end of a structured message.

So I propose we wrap stderr in a delegator instead. Essentially, any method invoked on `stderr` will still be redirect to it, but it gives us an opportunity to override `puts` and `print` to use structured logging, instead of just writing the strings directly to the pipe.

I tested this with the Tapioca add-on and I'm able to accurately see both progress notifications and messages in the output tab.

### DIFF
--- a/lib/ruby_lsp/ruby_lsp_rails/server.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/server.rb
@@ -3,6 +3,7 @@
 
 require "json"
 require "open3"
+require "delegate"
 
 module RubyLsp
   module Rails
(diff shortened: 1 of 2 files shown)


vinistock:
* **#564** <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp-rails/564?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/ruby-lsp-rails/564?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

andyw8:
Seeing some intermittent test failures in CI.

PR #570, opened 2025-02-03
Allow disabling pending migration prompt

### LINKED ISSUE #540: "Migrations are pending" popup is too aggressive
I often am running `bin/rails g migration SomeMigration` to get a blank migration and then writing it my editor instead of the generator. When I do this, the "Migrations are pending" popup immediately displays and asks me to run the (empty) migration.

From my perspective:
- I would like this popup to ignore empty migrations that I just created
- I would like the ability to disable this particular popup -- I wasn't able to find out how to do this in the list of "features enabled" in the main LSP

I think this has the potential for people to run the empty migration without realizing it ("I just click that button always") and then write the actual migration code and be confused why the changes don't seem to work (migration was marked as run so it doesnt re-run)

vinistock:
Closes #540

Allow disabling the pending migration prompt through an addon setting. To disable, users can add this to their configuration:

```json
"rubyLsp.addonSettings": {
   "Ruby LSP Rails": {
     "enablePendingMigrationsPrompt": false
   }
 }
```

### DIFF
--- a/lib/ruby_lsp/ruby_lsp_rails/addon.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/addon.rb
@@ -32,6 +32,12 @@ module RubyLsp
         @rails_runner_client = T.let(NullClient.new, RunnerClient)
         @global_state = T.let(nil, T.nilable(GlobalState))
         @outgoing_queue = T.let(nil, T.nilable(Thread::Queue))
+        @settings = T.let(
+          {
+            enablePendingMigrationsPrompt: true,
+          },
+          T::Hash[Symbol, T.untyped],
+        )
         @addon_mutex = T.let(Mutex.new, Mutex)
         @client_mutex = T.let(Mutex.new, Mutex)
         @client_mutex.lock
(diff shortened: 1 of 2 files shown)


lib/ruby_lsp/ruby_lsp_rails/addon.rb
```diff
@@ -32,6 +32,12 @@ def initialize
         @rails_runner_client = T.let(NullClient.new, RunnerClient)
         @global_state = T.let(nil, T.nilable(GlobalState))
         @outgoing_queue = T.let(nil, T.nilable(Thread::Queue))
+        @settings = T.let(
+          {
+            enable_pending_migration_prompt: true,
```
st0012, line 37: Should it be `enablePendingMigrationPrompt` instead? Right now config options like `rubyLsp.indexing`'s keys are all camelcases and I think that'd match most VS Code setting in general too.

andyw8: Also, I think it should be pluralized, to match the button shown in the browser.

PR #585, opened 2025-03-03
Test `server.rb` with better isolation

andyw8:
Since `server.rb` runs within the host Rails application, we should ensure we don't accidentally introduce something which depends on sorbet-runtime.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -10,14 +10,14 @@ load "rails/tasks/statistics.rake"
 require "bundler/gem_tasks"
 require "rake/testtask"
 
+# Since `server.rb` runs within the host Rails application, we want to ensure
+# we don't accidentally depend on sorbet-runtime. `server_test` intentionally doesn't use `test_helper`.
+# We run `server_test` in a seperate Rake task.
 Rake::TestTask.new(:test) do |t|
   t.libs << "test"
   t.libs << "lib"
   t.test_files = FileList["test/**/*_test.rb"] - ["test/ruby_lsp_rails/server_test.rb"]
 end
-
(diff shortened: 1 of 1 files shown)


sorbet/rbi/shims/server_test.rbi
```diff
@@ -0,0 +1,10 @@
+# typed: strict
+# frozen_string_literal: true
+
+module RubyLsp
+  module Rails
+    class Server
+      def print_it!; end
```
andyw8, line 7: Open to better ways to handle this.

andyw8: But then we need to load sorbet-runtime, which I am aiming to avoid.

vinistock: Ah, right...well, then there's no other way other than using the shim or waiting until we have comment support for unsafe.

Rakefile
```diff
@@ -13,7 +13,16 @@ require "rake/testtask"
...
+# Since `server.rb` runs within the host Rails application, we want to ensure
+# we don't accidentally depend on sorbet-runtime.
+
+Rake::TestTask.new(:server_test) do |t|
+  t.libs << "test"
+  t.libs << "lib"
+  t.test_files = ["test/ruby_lsp_rails/server_test.rb"]
+end
+
+task default: [:"db:setup", :test, :server_test]
```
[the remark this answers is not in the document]
andyw8: There are in the same process, but I suspect Rake is doing some cleanup of the load paths between jobs (I added a check to verify that `T` isn't defined). I'll look into a bit more.

PR #589, opened 2025-03-06
Fix tests for Rails main due to changes relating to default column value 

andyw8:
The default values behavior changed in https://github.com/rails/rails/pull/54683

### DIFF
--- a/test/ruby_lsp_rails/runner_client_test.rb
+++ b/test/ruby_lsp_rails/runner_client_test.rb
@@ -33,16 +33,30 @@ module RubyLsp
 
       test "#model returns information for the requested model" do
         # These columns are from the schema in the dummy app: test/dummy/db/schema.rb
-        columns = [
-          ["id", "integer", nil, false],
-          ["first_name", "string", "", true],
-          ["last_name", "string", nil, true],
-          ["age", "integer", "0", true],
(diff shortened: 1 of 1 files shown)


test/ruby_lsp_rails/runner_client_test.rb
```diff
@@ -33,16 +33,30 @@ class RunnerClientTest < ActiveSupport::TestCase
...
+            ["last_name", "string", nil, true],
+            ["age", "integer", "0", true],
+            ["created_at", "datetime", nil, false],
+            ["updated_at", "datetime", nil, false],
+            ["country_id", "integer", nil, false],
+            ["active", "boolean", "1", false],
+          ]
+        else
+          [
+            ["id", "integer", nil, false],
```
andyw8, line 50: There's some duplication but I think it's easier to follow this way.

vinistock:
Can you please explain what changed? There's no PR description and I'm confused as to why the default for an integer column would now be a string. Also, is that the only default that changed?

andyw8:
I had a comment in source, I've copied to the description.

>  I'm confused as to why the default for an integer column would now be a string.

Other way around. It's a string currently, the change in Rails makes it an integer.

vinistock (approved):

PR #591, opened 2025-04-07
Add test discovery support for ActiveSupport declarative syntax

alexcrocha:
### Motivation

Recent enhancements to the Ruby LSP introduced a new Test Discovery API. This PR leverages that API within the Ruby LSP Rails add-on to **add support for ActiveSupport's declarative test syntax** (`test "..." do ... end`).

While Ruby LSP supports various testing frameworks through different listeners (including standard Minitest and Spec styles), specific recognition for the Rails-centric declarative Minitest syntax was previously missing in the Ruby LSP Rails add-on. By incorporating support for this style, this change provides more complete test discovery and improves IDE integration for Rails projects, ensuring a seamless experience regardless of whether standard or declarative Minitest syntax is used.

### Implementation

- **Introduced `RubyLsp::Rails::RailsTestStyle`**: A new listener class inheriting from `RubyLsp::Listeners::TestDiscovery` was implemented to specifically handle Rails testing conventions.

- **Declarative Syntax Detection**: The listener identifies test classes employing the declarative style by looking for `ActiveSupport::Testing::Declarative` within the test class's ancestors. This ensures accurate detection even within complex inheritance chains.

- **Unified Test Parsing**: The implementation parses both standard `def test_` method definitions and `test "..." do` block structures, registering them appropriately as test items.

### DIFF
--- a/lib/ruby_lsp/ruby_lsp_rails/rails_test_style.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/rails_test_style.rb
@@ -43,7 +43,7 @@ module RubyLsp
         return unless first_arg.is_a?(Prism::StringNode)
 
         test_name = first_arg.content
-        return if test_name.empty?
+        test_name = "<empty test name>" if test_name.empty?
 
         add_test_item(node, test_name)
       end
--- a/test/ruby_lsp_rails/rails_test_style_test.rb
+++ b/test/ruby_lsp_rails/rails_test_style_test.rb
@@ -49,10 +49,10 @@ module RubyLsp
           assert_equal(1, items.length)
           test_class = items.first
           assert_equal("EmptyTest", test_class[:label])
-          assert_equal(1, test_class[:children].length)
+          assert_equal(2, test_class[:children].length)
 
           test_labels = test_class[:children].map { |i| i[:label] }
-          refute_includes(test_labels, "")
+          assert_includes(test_labels, "<empty test name>")
           assert_all_items_tagged_with(items, :rails)
         end
       end


lib/ruby_lsp/ruby_lsp_rails/rails_test_style.rb
```diff
@@ -0,0 +1,104 @@
+# typed: strict
+# frozen_string_literal: true
+
+module RubyLsp
+  module Rails
+    class RailsTestStyle < Listeners::TestDiscovery
+      extend T::Sig
```
vinistock, line 7: I don't think we need to `extend T::Sig` since we're only using comment signatures.

test/ruby_lsp_rails/rails_test_style_test.rb
```diff
@@ -0,0 +1,176 @@
...
+          assert_equal(2, test_class[:children].length)
+
+          test_labels = test_class[:children].map { |i| i[:label] }
+          assert_includes(test_labels, "first test")
+          assert_includes(test_labels, "second test")
+          assert_all_items_tagged_with(items, :rails)
+        end
+      end
+
+      test "discovers rails test with empty test name" do
```
vinistock, line 35: Does Rails prevent you from running a test with an empty name? I never tried it, so just confirming if we're choosing the right behaviour.

lib/ruby_lsp/ruby_lsp_rails/rails_test_style.rb
```diff
@@ -0,0 +1,104 @@
...
+
+      private
+
+      #: (Array[String] attached_ancestors, String fully_qualified_name) -> bool
+      def declarative_minitest?(attached_ancestors, fully_qualified_name)
+        return false unless attached_ancestors.include?("Minitest::Test")
+
+        # Check if this is a regular Minitest test class that doesn't use ActiveSupport's
+        # declarative testing syntax. We need to handle both ActiveSupport::TestCase tests
+        # and plain Minitest tests.
```
vinistock, line 71: I just tested this on the Rails add-on codebase itself and test classes actually don't inherit from `Minitest::Test`. We need to remove this early return and then everything works (and the comment no longer applies either).

```suggestion
        # The declarative test style is present as long as the class extends
        # ActiveSupport::Testing::Declarative
```

alexcrocha: Thanks for checking, I'll remove the early return.

I had pushed an amendment to the comment (will update again) but it's still stuck "processing updates" today 🤔 

<details><summary>Img</summary>
<p>

image


</p>
</details>

PR #625, opened 2025-06-06
Bump ruby-lsp to v0.24.0

alexcrocha:
Ruby-LSP v0.24.0 introduced breaking changes. This PR bumps to the newest release and resolves those breaking changes.

All tests pass. Manual testing shows that everything is working as intended.


lib/ruby_lsp/ruby_lsp_rails/rails_test_style.rb
```diff
@@ -21,7 +21,7 @@ def resolve_test_commands(items)
 
             children = item[:children]
             uri = URI(item[:uri])
-            path = uri.full_path
```
st0012, line 24: In ruby-lsp we can still use `#full_path` ([example](https://github.com/Shopify/ruby-lsp/blob/5a9ff9372bbdcdad4f8393f1480d174601d2cccf/lib/ruby_lsp/listeners/definition.rb#L319)). So I wonder if we the typechecking error is legit. Can you check if `uri.full_path` would raise an error in this case?

sorbet/rbi/shims/ruby_lsp.rbi
```diff
@@ -12,10 +12,9 @@ end
 module RubyLsp
   module Listeners
     class TestDiscovery
-      #: (ResponseBuilders::TestCollection response_builder, GlobalState global_state, Prism::Dispatcher dispatcher, URI::Generic uri) -> void
-      def initialize(response_builder, global_state, dispatcher, uri)
+      #: (ResponseBuilders::TestCollection response_builder, GlobalState global_state, URI::Generic uri) -> void
+      def initialize(response_builder, global_state, uri)
```
st0012, line 16: Can you test if we still need this shim?

lib/ruby_lsp/ruby_lsp_rails/rails_test_style.rb
alexcrocha: Somehow Sorbet did not recognize `full_path` alias anymore. We added a shim as a quick fix for now. We need to investigate why this happened.

sorbet/rbi/shims/ruby_lsp.rbi
alexcrocha: Double check and we still need it!

st0012 (approved):
Paired with @alexcrocha to walk through the changes 👍

PR #628, opened 2025-06-18
Rescue more exceptions in error handlers

KaanOzkan:
We were only rescuing StandardError's but we can also rescue some of the Exception class descendants


KaanOzkan:
> Do we need to specify each branch separately? Could we just group them all in the same rescue?

We could, I opted for consistency with other rescue's. I'll change it.

vinistock (approved):

PR #630, opened 2025-07-03
Add definition and hover support for `:through` associations

joshuay03:
Currently, cmd + clicking on the through association name goes to the same file as the association name. I've added definition support to navigate to the correct through association model file.

I've also added hover support as a follow-up to https://github.com/Shopify/ruby-lsp-rails/pull/616.

### DIFF
--- a/test/ruby_lsp_rails/definition_test.rb
+++ b/test/ruby_lsp_rails/definition_test.rb
@@ -53,6 +53,42 @@ module RubyLsp
         assert_equal(2, response[0].range.end.line)
       end
 
+      test "recognizes main association on has_many :through association" do
+        response = generate_definitions_for_source(<<~RUBY, { line: 2, character: 13 })
+          class Organization < ActiveRecord::Base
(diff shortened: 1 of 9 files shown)


test/ruby_lsp_rails/definition_test.rb
```diff
@@ -90,6 +116,32 @@ class User < ActiveRecord::Base
...
+        response = generate_definitions_for_source(<<~RUBY, { line: 4, character: 35 })
+          # typed: false
+
+          class User < ActiveRecord::Base
+            belongs_to :location, class_name: "Country"
+            has_one :country_flag, through: :location, source: :flag
+          end
+        RUBY
+
+        assert_equal(2, response.size)
```
joshuay03, line 129: Without patch:
Screenshot 2025-07-03 at 11 53 04 pm

vinistock (commented):
The Windows failures are unrelated. This makes sense, just added a question about the experience and the structure of the code

PR #633, opened 2025-07-17
Automatically name forked processes based on callers

vinistock:
From time to time, we have reports of extra child processes being created by the runtime server. It's hard to figure out where these are coming from because the user's application could be the one creating them. It could also be Tapioca or some other add-on.

This PR patches `fork` to always name the forked process using the file path of the caller, so that we know where fork was called. While `caller_locations` is a bit expensive, I think it's worth the extra help in debugging. Please let me know if you agree.

### DIFF
--- a/lib/ruby_lsp/ruby_lsp_rails/server.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/server.rb
@@ -542,6 +542,25 @@ module RubyLsp
   end
 end
 
+# Patch fork to name processes based on the caller's file path. This is useful for figuring out what is creating more
+# child processes from the runtime server, so that we can optimize and more easily debug orphaned processes
+# @requires_ancestor: Kernel
+module ForkHook
+  #: (*untyped) -> Integer?
+  def _fork(*args)
+    pid = super
+
+    if pid == 0
+      fork_caller = caller_locations(1, 1)&.first
+      Process.setproctitle("ruby-lsp-rails: #{fork_caller.path}") if fork_caller
+    end
+
+    pid
+  end
+
+  Process.singleton_class.prepend(self)
+end
+
 if ARGV.first == "start"
   RubyLsp::Rails::Server.new(capabilities: JSON.parse(ARGV[1], symbolize_names: true)).start
 end
--- a/test/ruby_lsp_rails/server_test.rb
+++ b/test/ruby_lsp_rails/server_test.rb
@@ -8,6 +8,8 @@ class ServerTest < ActiveSupport::TestCase
   setup do
     @stdout = StringIO.new
     @stderr = StringIO.new
+    RubyLsp::Rails::ServerAddon.instance_variable_set(:@server_addon_classes, [])
+    RubyLsp::Rails::ServerAddon.instance_variable_set(:@server_addons, {})
     @server = RubyLsp::Rails::Server.new(stdout: @stdout, stderr: @stderr, override_default_output_device: false)
   end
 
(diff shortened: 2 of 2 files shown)


test/ruby_lsp_rails/server_test.rb
```diff
@@ -268,6 +268,38 @@ def print_it!
...
+        def execute(request, params)
+          pid = fork do
+            # We can't directly send a message in theses tests because we're using a StringIO as stdout instead of the
+            # actual pipe, which means that the child process doesn't have access to the same object
+            File.write("process_name.txt", $0)
+          end
+
+          Process.wait(pid)
+          send_message({ process_name: File.read("process_name.txt") })
+          File.delete("process_name.txt")
```
Morriar, line 289: Let's make sure this gets cleaned?

```suggestion
        ensure
          File.delete("process_name.txt")
```

lib/ruby_lsp/ruby_lsp_rails/server.rb
```diff
@@ -529,6 +529,23 @@ def database_supports_indexing?(model)
...
 
+# Patch fork to name processes based on the caller's file path. This is useful for figuring out what is creating more
+# child processes from the runtime server, so that we can optimize and more easily debug orphaned processes
+# @requires_ancestor: Kernel
+module ForkPatch
+  #: () { () -> void } -> Integer?
+  def fork(&block)
+    super do
+      fork_caller = caller_locations(3, 1)&.first
+      $0 = "ruby-lsp-rails: #{fork_caller&.path}"
```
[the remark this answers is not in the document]
vinistock: `Process.setproctitle` is actually better as it changes the title without changing `$0`.

lib/ruby_lsp/ruby_lsp_rails/server.rb
```diff
@@ -529,6 +529,32 @@ def database_supports_indexing?(model)
   end
 end
 
+# Patch fork to name processes based on the caller's file path. This is useful for figuring out what is creating more
+# child processes from the runtime server, so that we can optimize and more easily debug orphaned processes
+# @requires_ancestor: Kernel
+module ForkPatch
+  #: () ?{ () -> void } -> Integer?
+  def fork(&block)
```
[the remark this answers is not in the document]
vinistock: Nice

lib/ruby_lsp/ruby_lsp_rails/server.rb
```diff
@@ -529,6 +529,22 @@ def database_supports_indexing?(model)
...
+# child processes from the runtime server, so that we can optimize and more easily debug orphaned processes
+# @requires_ancestor: Kernel
+module ForkHook
+  #: (*untyped) -> Integer?
+  def _fork(*args)
+    fork_caller = caller_locations(1, 1)&.first
+
+    pid = super
+    Process.setproctitle("ruby-lsp-rails: #{fork_caller.path}") if fork_caller && pid == 0
+    pid
```
paracycle, line 542: Small optimization here:
```suggestion
    pid = super
    
    if pid == 0
      fork_caller = caller_locations(1, 1)&.first

      Process.setproctitle("ruby-lsp-rails: #{fork_caller.path}") if fork_caller
    end
    
    pid
```

so that we only call `caller_locations` if we wanted to set the process title.

test/ruby_lsp_rails/server_test.rb
```diff
@@ -268,6 +270,79 @@ def print_it!
...
+    File.write(addon_path, <<~RUBY)
+      class MyServerAddon < RubyLsp::Rails::ServerAddon
+        def name
+          "MyAddon"
+        end
+
+        def execute(request, params)
+          file = "process_name.txt"
+          pid = fork do
+            # We can't directly send a message in theses tests because we're using a StringIO as stdout instead of the
```
paracycle, line 286:
```suggestion
            # We can't directly send a message in these tests because we're using a StringIO as stdout instead of the
```

test/ruby_lsp_rails/server_test.rb
```diff
@@ -268,6 +270,79 @@ def print_it!
...
+        end
+
+        def execute(request, params)
+          file = "other_process_name.txt"
+          pid = fork
+
+          if pid
+            Process.wait(pid)
+            send_message({ process_name: File.read(file) })
+            File.delete(file)
```
paracycle, line 325: Should we also assert that this process' title (the parent process) was not changed after fork?

test/ruby_lsp_rails/server_test.rb
```diff
@@ -268,6 +270,79 @@ def print_it!
...
+        def execute(request, params)
+          file = "process_name.txt"
+          pid = fork do
+            # We can't directly send a message in theses tests because we're using a StringIO as stdout instead of the
+            # actual pipe, which means that the child process doesn't have access to the same object
+            process_title = `ps -p \#{Process.pid} -o comm=`.lines.last.strip
+            File.write(file, process_title)
+          end
+
+          Process.wait(pid)
```
paracycle, line 292: Should we assert that this process' title (the parent fork) wasn't changed by the fork?

test/ruby_lsp_rails/server_test.rb
vinistock: Yes, good idea. I added code to assert that it didn't change.

test/ruby_lsp_rails/server_test.rb
vinistock: Yes, good idea. I added code to assert that it didn't change.

PR #638, opened 2025-07-24
Add go to definition in validation conditionals

domingo2000:
In [ActiveRecord::Validations](https://api.rubyonrails.org/classes/ActiveModel/Validations/ClassMethods.html#method-i-validates) API there are multiple class methods that can take the `if` and `unless` parameter with a symbol callback. It would be nice to have go to definition in those symbols and this PR solves that.

The implementation takes all the cases from the documentation and provides the go to definitions for that cases.

[Screencast from 2025-07-24 00-41-09.webm](https://github.com/user-attachments/assets/ab70bef5-04a2-487d-8d42-f255489fd376)


domingo2000:
@vinistock I created the PR without issue because is very similar to previous PR. Happy to receive feedback on this.

domingo2000:
> I tested this and I noticed an existing bug that also exists on this PR. Callbacks can also be conditional with `if` or `unless`.
> 
> Right now, if you go to definition on any part of the callback, it will always take you to the method and not the conditional. This PR has the reverse bug, where clicking on the validation method always takes you to the `if` or `unless` conditional.
> 
> For example:
> 
> ```ruby
> validate :something, if: :should_do_it?
>          # ^^^ jumping here should take you to `something` and not `should_do_it?`
> 
> # Callbacks right now are the reverse
> 
> before_save :do_something, if: :should_do_it?
>                                # ^^^ jumping here takes you to `do_something`
> ```
> 
> It might be the case that we can fix both the behaviour for callbacks and for validations with the same logic, given that it's basically the same case of jumping to symbols. Can we please address this as part of this PR?

@vinistock  good catch, the whole call_node context of the symbol was handled, so the responses got duplications or definitions that where not requested. I fixed the callbacks and if/unless conditional duplications. So the response only contains the go to definition for the callback/conditional method requested.

Also resolved all comments 👍🏻 .

domingo2000:
The scope of this PR is just to handle the if/unless conditionals of callbacks and validations. So i do not include the go to definition for the validations first argument, y can make a follow up with that aditional feature.

lib/ruby_lsp/ruby_lsp_rails/definition.rb
```diff
@@ -51,53 +51,92 @@ def on_string_node_enter(node)
...
+        return unless call_node
 
-        message = node.message
+        arguments = call_node.arguments&.arguments
+        return unless arguments
 
-        return unless message
+        focus_argument = arguments.find do |argument|
+          true if argument == node
+        end
```
vinistock, line 100:
```suggestion
        focus_argument = arguments.find { |argument| argument == node }
```

lib/ruby_lsp/ruby_lsp_rails/definition.rb
```diff
@@ -51,53 +51,92 @@ def on_string_node_enter(node)
...
+        return unless focus_argument
 
-        arguments.each do |argument|
-          name = case argument
-          when Prism::SymbolNode
-            argument.value
-          when Prism::StringNode
-            argument.content
-          end
+        return unless node.is_a?(Prism::SymbolNode)
```
vinistock, line 130: Do validations not accept a string argument?

lib/ruby_lsp/ruby_lsp_rails/definition.rb
```diff
@@ -51,53 +51,92 @@ def on_string_node_enter(node)
...
 
-          next unless name
+        name = node.value
+        return unless name
 
+        case message
+        when "validates_with"
+          # validates_with uses constants, not symbols - skip (handled by constant resolution)
+          nil
+        else
```
vinistock, line 139: Maybe we can simply do

```ruby
return if message == "validates_with"
```

lib/ruby_lsp/ruby_lsp_rails/definition.rb
```diff
@@ -141,6 +180,42 @@ def collect_definitions(name)
...
         end
       end
+
+      #: ((Prism::SymbolNode | Prism::StringNode) node) -> void
+      def handle_if_unless_conditional(node)
+        call_node = @node_context.call_node
+        return unless call_node
+
+        arguments = call_node.arguments&.arguments
+        return unless arguments
```
vinistock, line 190: It looks like most of these methods need to know if there's a surrounding call_node and if there are any arguments.

How about we pull this into `on_call_node_enter` and then pass whatever is necessary through parameters to these methods, so that we can avoid retrieving the arguments in each one of these?

vinistock (commented):
Fantastic! Thank you for fixing the existing bugs. It works really well now.

I left some styling comments, but nothing major. The changes look good.

lib/ruby_lsp/ruby_lsp_rails/definition.rb
domingo2000: I implemented only for symbols because, even though strings maybe work, through all the documentation and rails API docs all the examples are given with symbols, so maybe some parts work with strings _and_ symbols. But i prefer to enforce symbol usage, this is closer to what the Rails guides and API docs shows. With that we are sure that the go to definition are correct because the symbols are widely accepted and documented, but the string don't.

lib/ruby_lsp/ruby_lsp_rails/definition.rb
domingo2000: It cannot be put in the on_call_node_enter and _then_ pass the parameters to the methods, because this methods needs to be called on symbol node enter to pass the symbol and search for the focused symbol.

But I refactored to deduplicate that code, because all DSL needs argument in their call_node context.

Issue #639, opened 2025-07-27
Add support for I18n


domingo2000:
# Provide I18n Support

It would be great to have [I18n](https://guides.rubyonrails.org/i18n.html) translation support in the Rails addon. The main use cases when implementing internationalization with I18n are:

### Use Cases

1. Read what the translation says. e.g., `I18n.t("hello")` returns `en: "Hello World", es: "¡Hola Mundo!"`.
2. Find out if a translation is missing.
3. Locate where a translation is defined.
4. Create new translations.
5. Update translations (for a key, across one or multiple locales).

### Idea

The goal of this issue (if you agree) is to build a simple implementation that addresses the use cases above, avoiding and solving some of the drawbacks we’ve encountered over the years.

Starting with point 1 in the list:

- [ ] Provide hover support for translations in `I18n.t("key")` calls.
- [ ] Synchronize the I18n backend when users modify their translation files.

I will open a PR showcasing an initial implementation.

Next, we could work on:

- [ ] Autocompletion while typing, e.g., `I18n.t("hel`
- [ ] “Go to definition” for translations

These features are already implemented in our custom solution.

### Some Limitations with the Rails I18n API

In our implementation, we store the file path for each translation key. However, the Rails I18n API doesn’t store the source location of translations, which makes points 3, 4, and 5 difficult to implement without modifying the I18n backend.

That said, points 1 and 2 already provide substantial value on their own.

### Current Addon Implementation

At our organization, we’ve been using [ruby-lsp-i18n](https://github.com/bukhr/ruby-lsp-i18n) internally for over a year. It addresses all of the use cases listed above and has significantly improved the developer experience when working with translations.

Because a project can contain millions of translations, managing them efficiently is crucial. For example, to accomplish point 1, developers often resort to opening a Rails console just to call `I18n.t`.

Here's a summary of what our addon currently does:

- On startup, it indexes all translation keys across all YAML files.
- It creates an index: `key => (value, file)`.
- It builds a Trie from the values to enable autocompletion.
- It displays an inlay hint for the value of the base language (`es`) and shows a markdown tooltip with:
  - A link to the file (serving as a "go to definition").
  - A “recommended” file path for new translations (based on our internal convention).
- It attaches a file watcher to keep all translations in sync when files change — updating both the index and the Trie on each change.

### Current Drawbacks of Our Implementation

[...]

PR #640, opened 2025-07-27
Add hover on i18n translations

domingo2000:
First implementation for I18n support (#639)

## How it works

1. Attach file watcher for all translation files
2. Add server request to reload translations on file changes. The reload is fast because does not need to reload the entire rails server, just the I18n.
3. Add server request to get I18n translations for all locales
4. On hover of `I18n.t("hello")` key string argument, request the server all the translations available and provide hover info for that.

It address points 1 and 2 of the issue list:

1. Read what the translation says. e.g., I18n.t("hello") returns en: "Hello World", es: "¡Hola Mundo!".
2. Find out if a translation is missing.

And is usefull for updating and creating translations because it gives feedback that the used translations "is correct" because makes direct calls to the I18n backend while typing and hovering it. Warning if the translation is missing.

This gives the following experience: 
[Screencast from 2025-07-27 00-40-33.webm](https://github.com/user-attachments/assets/8efa6f59-f5b6-4006-b1f5-6e75c32f2f15)

### DIFF
--- a/lib/ruby_lsp/ruby_lsp_rails/server.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/server.rb
@@ -332,6 +332,17 @@ module RubyLsp
           with_request_error_handling(request) do
             send_result(resolve_route_info(params))
           end
+        when "i18n"
+          with_request_error_handling(request) do
+            result = resolve_i18n_key(params.fetch(:key))
+            send_result(result)
+          end
+        when "reload_i18n"
+          with_progress("rails-reload-i18n", "Reloading Ruby LSP Rails I18n") do
+            with_notification_error_handling(request) do
+              I18n.reload! if defined?(I18n) && I18n.respond_to?(:reload!)
+            end
+          end
         when "server_addon/register"
           with_notification_error_handling(request) do
             require params[:server_addon_path]
@@ -524,6 +535,16 @@ module RubyLsp
         @database_supports_indexing = true
       rescue NotImplementedError
         @database_supports_indexing = false
+      end
+
+      #: (String) -> Hash[Symbol, String]
+      def resolve_i18n_key(key)
+        locales = I18n.available_locales
+        result = {}
+        locales.each.map do |locale|
+          result[locale] = I18n.t(key, locale: locale, default: "⚠️ translation missing")
+        end
+        result
       end
     end
   end
--- a/test/ruby_lsp_rails/hover_test.rb
+++ b/test/ruby_lsp_rails/hover_test.rb
@@ -321,6 +321,28 @@ module RubyLsp
         CONTENT
       end
 
+      test "returns I18n translation information" do
+        expected_response = {
+          en: "hello",
+          es: "hola",
+          fr: "bonjour",
+        }
+
+        RunnerClient.any_instance.stubs(i18n: expected_response)
+
+        response = hover_on_source(<<~RUBY, { line: 0, character: 9 })
+          I18n.t("foo")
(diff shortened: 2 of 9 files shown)


lib/ruby_lsp/ruby_lsp_rails/server.rb
```diff
@@ -332,6 +332,17 @@ def execute(request, params)
...
           end
+        when "i18n"
+          with_request_error_handling(request) do
+            result = resolve_i18n_key(params.fetch(:key))
+            send_result(result)
+          end
+        when "reload_i18n"
+          with_progress("rails-reload-i18n", "Reloading Ruby LSP Rails I18n") do
+            with_notification_error_handling(request) do
+              I18n.reload! if defined?(I18n) && I18n.respond_to?(:reload!)
```
domingo2000, line 343: This is provided because other I18n backend maybe does not implement the reloading feature.

test/ruby_lsp_rails/hover_test.rb
```diff
@@ -321,6 +321,28 @@ class Bar < ApplicationRecord
...
+        expected_response = {
+          en: "hello",
+          es: "hola",
+          fr: "bonjour",
+        }
+
+        RunnerClient.any_instance.stubs(i18n: expected_response)
+
+        response = hover_on_source(<<~RUBY, { line: 0, character: 9 })
+          I18n.t("foo")
```
brunoprietog, line 334: This could also be just `t` if we consider that you could be inside Action View, but perhaps that is beyond the scope.

domingo2000: Yes, its out of scope. This changes only support I18n.t("some-key") calls. With absolute key and string literal as argument.

To support `t` notation for ActionView, we need the context of which file is beign parsed and be able to complete the call of I18n in an ActionView context, or with the scope of that specific file. It could be a nice follow up if you want to help 👍🏻 .

janko: Answering myself: we'd need to resolve the controller path for sure, which isn't trivial if one wants to handle custom `view_paths`. I implemented it [here](https://github.com/Shopify/ruby-lsp-rails/pull/659/files#diff-06811bf1a3f210c0c6cfdcce305b6487c2223f468bcd1346f26ebd236fcec6bbR290-R310), so at least it could be reused. But I agree it's for a follow-up PR 👍🏻

Drowze:
This is a fantastic addition to `ruby-lsp-rails` - hope it gets reviewed & merged soon 🤞

andy-rootly:
It would also be nice if Go To Definition jumped to the entry in the locale file.

lib/ruby_lsp/ruby_lsp_rails/server.rb
```diff
@@ -525,6 +536,16 @@ def database_supports_indexing?(model)
...
       end
+
+      #: (String) -> Hash[Symbol, String]
+      def resolve_i18n_key(key)
+        locales = I18n.available_locales
+        result = {}
+        locales.each.map do |locale|
+          result[locale] = I18n.t(key, locale: locale, default: "⚠️ translation missing")
+        end
+        result
```
rafaelfranca:
```suggestion
        locales.each_with_object do |locale, result|
          result[locale] = I18n.t(key, locale: locale, default: "⚠️ translation missing")
        end
```

rafaelfranca (commented):
A few small thinks

rafaelfranca (approved):

PR #642, opened 2025-08-11
Use unformatted descriptions as test example labels

vinistock:
We were using the formatted example descriptions as both the ID and the label. That's not the best because it doesn't match what the user wrote (with spaces and as plain text) and also ends up requiring folks to add underscores for filtering examples.

This PR starts using the unformatted names as labels, which better matches the structure of the file. Note that labels are only used for presentation and filtering in the UI, we rely on the IDs for the actual running functionality.

### DIFF
--- a/lib/ruby_lsp/ruby_lsp_rails/rails_test_style.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/rails_test_style.rb
@@ -67,9 +67,11 @@ module RubyLsp
       def on_class_node_enter(node)
         with_test_ancestor_tracking(node) do |name, ancestors|
           if declarative_minitest?(ancestors, name)
+            label = constant_name(node.constant_path) || name_with_dynamic_reference(node.constant_path)
+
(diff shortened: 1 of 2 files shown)


lib/ruby_lsp/ruby_lsp_rails/rails_test_style.rb
```diff
@@ -67,9 +67,12 @@ def initialize(response_builder, global_state, dispatcher, uri)
       def on_class_node_enter(node)
         with_test_ancestor_tracking(node) do |name, ancestors|
           if declarative_minitest?(ancestors, name)
+            label = constant_name(node.constant_path)
+            label ||= name_with_dynamic_reference(node.constant_path)
```
Morriar, line 71: nit

```suggestion
            label = constant_name(node.constant_path) || name_with_dynamic_reference(node.constant_path)
```

Morriar (approved):

PR #697, opened 2026-03-24
Use Prism parser for Sorbet

Morriar:
Adds `--parser=prism` to `sorbet/config` to use the Prism parser instead of the legacy Whitequark parser.


vinistock (approved):
We may need to upgrade Sorbet too

PR #708, opened 2026-04-27
Ensure runtime server exits after shutdown

vinistock:
A lot of our life cycle was getting really clunky and we are frequently seeing tests hang, especially on Windows. This PR aims to improve the situation:

1. Ensure that we `join` the boot thread when deactivating the add-on. This one is mostly about tests. Tests quickly create new add-on instances, which boots the Rails app. If we don't join the boot thread, we risk finishing the tests before the Rails app is up and we may end up with orphaned processes
2. Got rid of the `force_kill` approach. This was basically a brute force way of trying to ensure things were shutting down properly
3. Started making sure that the runtime server calls `exit!(0)` once the main loop finishes, to terminate child processes with it
4. Removed the process group ID stuff, which I don't believe was actually doing anything useful
5. Started handling early termination of stdin in the runtime server, so that it shuts down gracefully if the pipe was closed on the other end

### DIFF
--- a/lib/ruby_lsp/ruby_lsp_rails/addon.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/addon.rb
@@ -39,7 +39,7 @@ module RubyLsp
         @client_mutex = Mutex.new #: Mutex
         @client_mutex.lock
 
-        Thread.new do
+        @boot_thread = Thread.new do
           @addon_mutex.synchronize do
             # We need to ensure the Rails client is fully loaded before we activate the server addons
             @client_mutex.synchronize do
@@ -50,7 +50,7 @@ module RubyLsp
             end
             offer_to_run_pending_migrations
           end
-        end
+        end #: Thread
       end
 
       #: -> RunnerClient
@@ -77,6 +77,7 @@ module RubyLsp
       # @override
       #: -> void
       def deactivate
+        @boot_thread.join
         @rails_runner_client.shutdown
       end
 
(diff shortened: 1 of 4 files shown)


vinistock:
* **#708** <a href="https://app.graphite.com/github/pr/Shopify/ruby-lsp-rails/708?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/ruby-lsp-rails/708?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lib/ruby_lsp/ruby_lsp_rails/addon.rb
```diff
@@ -78,6 +78,7 @@ def activate(global_state, outgoing_queue)
       # @override
       #: -> void
       def deactivate
+        @boot_thread.join
```
st0012, line 80: Should this be `@boot_thread&.join` in case `@client_mutex` is never acquired before `deactivate` is called?

st0012 (approved):

lib/ruby_lsp/ruby_lsp_rails/addon.rb
vinistock: That should never happen. We eagerly instantiate all add-on instances, then invoke `activate` on the instance and only after that can `deactivate` be called. Even with multiple threads, I don't believe this scenario can happen.

PR #727, opened 2026-08-04
Fix intermittent server boot failure caused by working directory race

soutaro:
### Motivation

CI hangs intermittently on Ruby 4.0.6 (first seen in #724 / #725) — jobs time out after 25 minutes with the test suite stuck mid-run. The cause is a working-directory race in `RunnerClient.create_client`, not setup-ruby or gem updates.

`create_client` checks `File.exist?("bin/rails")` relative to the process working directory, and spawns the `rails runner` server with the inherited cwd. But the working directory is process-global state that other threads mutate: RuboCop's config loader evaluates configs with `Dir.chdir(config dir) { ERB.new(...).result }` while resolving `inherit_gem`. When the Ruby LSP RuboCop addon activates concurrently with the Rails addon boot thread, the boot thread can observe the `rubocop-shopify` gem directory as cwd. The `bin/rails` check then fails and `create_client` silently falls back to a `NullClient`.

The race is long-standing but almost never fired until Ruby 4.0.6, whose thread-scheduler changes ([Bug #21685](https://bugs.ruby-lang.org/issues/21685) backport) altered thread interleaving enough to make it fire several times per 100 boots on CI runners. That's why bumping setup-ruby past 1.317.0 (which resolves `"4.0"` to 4.0.6 instead of 4.0.5) broke CI. Trace from a failing boot captured in the investigation PR (#726, now closed):

```
[RuboCop] Activating RuboCop LSP addon 1.88.2.
Ruby LSP Rails create_client start (cwd=.../vendor/bundle/ruby/4.0.0/gems/rubocop-shopify-2.18.0)
Ruby LSP Rails create_client: bin/rails NOT found (cwd=.../gems/rubocop-shopify-2.18.0)
[RuboCop] Initialized RuboCop LSP addon 1.88.2.
```

Note that the hang itself only happens in this repo's test suite: several tests busy-wait with `sleep(0.1) while ...NullClient` for the real client to boot, and the loop never terminates when boot fails because `NullClient` is the final state. In real editor usage a boot failure does not hang anything — the add-on falls back to the `NullClient`, logs a warning, and Rails-dependent features are silently unavailable.

### Implementation

- Resolve `bin/rails` against `global_state.workspace_path` instead of the cwd, and spawn the server with an explicit `chdir:` to the workspace path, so neither depends on racy process-global state.
- Fix the tests' unbounded busy-wait as well: add `Addon#join_boot_thread` so the boot outcome can be awaited deterministically, and make the tests join the boot thread and assert that the client actually booted. A boot failure now fails the test immediately with a clear message instead of hanging the suite until the CI timeout.

### DIFF
--- a/lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
@@ -10,13 +10,15 @@ module RubyLsp
       class << self
         #: (Thread::Queue outgoing_queue, RubyLsp::GlobalState global_state) -> RunnerClient
         def create_client(outgoing_queue, global_state)
-          if File.exist?("bin/rails")
+          workspace_path = global_state.workspace_path
+
+          if File.exist?(File.join(workspace_path, "bin", "rails"))
             new(outgoing_queue, global_state)
           else
             unless outgoing_queue.closed?
               outgoing_queue << RubyLsp::Notification.window_log_message(
                 <<~MESSAGE.chomp,
-                  Ruby LSP Rails failed to locate bin/rails in the current directory: #{Dir.pwd}
+                  Ruby LSP Rails failed to locate bin/rails in the workspace: #{workspace_path}
                   Server dependent features will not be available
                 MESSAGE
                 type: RubyLsp::Constant::MessageType::WARNING,
@@ -64,6 +66,7 @@ module RubyLsp
             "#{__dir__}/server.rb",
             "start",
             server_relevant_capabilities(global_state),
+            { chdir: global_state.workspace_path },
           )
         end
 
--- a/lib/ruby_lsp/ruby_lsp_rails/addon.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/addon.rb
@@ -58,6 +58,13 @@ module RubyLsp
         @addon_mutex.synchronize { @rails_runner_client }
       end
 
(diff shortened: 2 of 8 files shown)


st0012 (approved):

lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
```diff
@@ -10,13 +10,15 @@ class RunnerClient
       class << self
         #: (Thread::Queue outgoing_queue, RubyLsp::GlobalState global_state) -> RunnerClient
         def create_client(outgoing_queue, global_state)
-          if File.exist?("bin/rails")
+          workspace_path = global_state.workspace_path
+
+          if File.exist?(File.join(workspace_path, "bin", "rails"))
```
Morriar, line 15: Do we have a test that tries to create a client from outside the `workspace_path`?

Morriar (approved):

jesse-shopify (approved):
I really like seeing the removal of all the `sleep` calls. This is a much improved design.

lib/ruby_lsp/ruby_lsp_rails/runner_client.rb
soutaro: Not before — added one in 35edca5. It creates a client while the process working directory is a temporary directory outside the workspace, and asserts a real client boots and can answer a request.

PR #730, opened 2026-09-05
Link schema hover to the model's table definition

### LINKED ISSUE #52: Enhance our schema hover link to jump to the right table
Currently, our schema link just points to the file. However, we can link to the specific table for the model, which is a richer experience.

One possibility of how we can do this is by
1. Start returning the table name from our middleware
2. During the `activate` call parse `schema.rb` and create a hash of table names to locations
3. When executing hover, search the hash for the right line to jump to

**Concerns**

- If there are modifications in schema.rb, we would need to rebuild the hash. For that, we need [file watching](https://github.com/Shopify/ruby-lsp/pull/632) to be merged in the Ruby LSP and then we have to design a way in which extensions can ask the Ruby LSP to notify them when certain files are changed.
- This won't work for `structure.sql`. We probably don't want a SQL parser, but we might be able to achieve something with regexes

yoshiyoshiharu:
Fixes #52.

The `Schema` link in a model's hover opens `db/schema.rb` at the top, so you still have to find the table yourself. This makes it jump to the model's `create_table` block.

This was already being worked on in #212, which takes a similar approach.
That PR is still open but has had no activity since March 2024, so rather than leave the issue sitting I put together a fresh implementation. Happy to close this in favour of #212 if it's still moving.

### DIFF
--- a/lib/ruby_lsp/ruby_lsp_rails/support/schema_table_location_visitor.rb
+++ b/lib/ruby_lsp/ruby_lsp_rails/support/schema_table_location_visitor.rb
@@ -1,0 +1,62 @@
+# typed: strict
+# frozen_string_literal: true
+
+module RubyLsp
+  module Rails
+    module Support
+      class SchemaTableLocationVisitor < Prism::Visitor
+        class << self
+          #: (String source, String table_name) -> Prism::Location?
+          def find(source, table_name)
+            table_locations(source)[table_name]
+          end
+
+          #: -> void
+          def expire_cache
+            @cache.clear
+          end
+
+          private
+
+          #: (String source) -> Hash[String, Prism::Location]
+          def table_locations(source)
+            @cache[source] ||= begin
+              visitor = new
+              Prism.parse(source).value.accept(visitor)
+              visitor.tables
+            end
+          end
+        end
+
+        @cache = {} #: Hash[String, Hash[String, Prism::Location]]
+
+        #: -> void
+        def initialize
+          super
+          @tables = {} #: Hash[String, Prism::Location]
+        end
+
+        #: Hash[String, Prism::Location]
+        attr_reader :tables
+
+        #: (Prism::CallNode node) -> void
+        def visit_call_node(node)
+          if node.message == "create_table"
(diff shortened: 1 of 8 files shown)


yoshiyoshiharu:
I have signed the CLA!

lib/ruby_lsp/ruby_lsp_rails/support/schema_table_location_visitor.rb
```diff
@@ -0,0 +1,62 @@
...
+            when Prism::StringNode
+              first_argument.unescaped
+            when Prism::SymbolNode
+              first_argument.unescaped
+            end
+
+            @tables[name] = node.location if name
+          end
+
+          super
```
soutaro, line 57: Could we stop traversing once we've recorded a `create_table` location? We don't need to visit the column definitions inside the block.

lib/ruby_lsp/ruby_lsp_rails/support/schema_table_location_visitor.rb
```diff
@@ -0,0 +1,62 @@
+# typed: strict
+# frozen_string_literal: true
+
+module RubyLsp
+  module Rails
+    module Support
+      class SchemaTableLocationVisitor < Prism::Visitor
+        class << self
+          #: (String source, String table_name) -> Prism::Location?
+          def find(source, table_name)
```
soutaro, line 10:
```suggestion
          def find(source, table_name, base_prefix:, base_suffix:)
```

Could we handle `ActiveRecord::Base.table_name_prefix` and `table_name_suffix` here? With `app_` and `_records` configured, `User.table_name` returns `app_users_records`, but the schema contains `create_table "users"`, so this lookup misses the table.

Note that model-specific prefixes and suffixes can remain in `schema.rb`. Maybe we could try the given `table_name` first and, if no match is found, retry with the Base prefix/suffix removed.

test/ruby_lsp_rails/hover_test.rb
soutaro, line 260: Could you add `table_name: "users"` to `expected_response` here so we can verify that the `structure.sql` fallback still works when `table_name` is present?

